PR #65, opened 2026-04-02
added AMD ROCm for mi300

robmsmt:


AryanAhadinia (approved):
Great to see that! We should also consider using AWS Libfabric for the multi-node models.

PR #71, opened 2026-04-09
Use FirecREST v2

AryanAhadinia:


robmsmt (approved):

AryanAhadinia:
Well, we were already using v2 client, I thought that we should also switch to v2 credentials, but apparently, the credentials classes are solely in v1. So, all it good and no need to change.

PR #78, opened 2026-04-10
Early Apertus 1.5 Support

AryanAhadinia:
To be tested and enhanced when the cluster is up.

PR #95, opened 2026-04-22
Add Qwen3.6-27B (vllm DP=2 TP=2)

Alvorecer721:
## Summary
- Register **Qwen/Qwen3.6-27B** as a preconfigured model: vllm with `--data-parallel-size 2 --tensor-parallel-size 2 --max-model-len 65536 --reasoning-parser qwen3` on one GH200 node.
- Two template fixes uncovered while getting the launch working; both are independent of Qwen3.6 and benefit every model.

### 1. `models.json` — new entry
Qwen3.6-27B is a 27B dense model (Apache 2.0). On 4×GH200 the cleanest layout is DP=2 × TP=2 inside a single `vllm serve` process — one endpoint, no router needed. Model path resolves via the standard registry symlink at `/capstor/store/cscs/swissai/infra01/hf_models/models/Qwen/Qwen3.6-27B`.

### 2. `template.jinja` — unset inherited pyxis env (bug fix)
```bash
unset $(env | awk -F= '/^SLURM_SPANK__SLURM_SPANK_OPTION_pyxis_/ {print $1}') 2>/dev/null || true
unset SBATCH_ENVIRONMENT SRUN_ENVIRONMENT
```
When `sml` is invoked from a nested pyxis shell (e.g. VSCode `code` session, or any shell that was started with `--environment`), those vars leak into `sbatch` and every subsequent `srun --environment ...` errors with *"--environment specified multiple times"*. Clearing them at the top of the batch script makes the explicit `--environment` on each srun the single source of truth.

### 3. `template.jinja` — router health check uses full worker URL (bug fix)
```diff
-for worker_ip in ${worker_head_ips[*]}; do
-    while [ "$(curl … http://${worker_ip}:${WORKER_PORT}/health)" != "200" ]; do
+for worker_url in ${worker_urls[*]}; do
+    while [ "$(curl … ${worker_url}/health)" != "200" ]; do
```
The old form hardcoded `WORKER_PORT` (5000) for every worker, so workers bound to alternate ports on the same host were never polled correctly. `worker_urls` already carries the right `host:port`.

## Test plan
- [x] `sml preconfigured --model Qwen/Qwen3.6-27B --framework vllm --time 12:00:00` submits cleanly
- [x] vllm loads weights and `/v1/chat/completions` responds (endpoint on :8080, DP=2/TP=2 verified via process layout)
- [x] Launch succeeds when submitted from a pyxis-active shell (previously failed with "--environment specified multiple times")
- [ ] Existing multi-worker sglang launches still work (template change is a strict fix — polling `worker_urls[*]` replaces polling `${worker_ip}:${WORKER_PORT}`; `worker_urls` was already populated)


src/swiss_ai_model_launch/assets/template.jinja
```diff
@@ -238,12 +242,12 @@ set -ex
...
-for worker_ip in ${worker_head_ips[*]}; do
-    echo \"Checking worker at \$worker_ip...\"
-    while [ \"\$(curl --noproxy \"*\" -s -o /dev/null -w '%{http_code}' http://\${worker_ip}:${WORKER_PORT}/health)\" != \"200\" ]; do
+for worker_url in ${worker_urls[*]}; do
+    echo \"Checking worker at \$worker_url...\"
+    while [ \"\$(curl --noproxy \"*\" -s -o /dev/null -w '%{http_code}' \${worker_url}/health)\" != \"200\" ]; do
         sleep 10
     done
-    echo \"Worker \$worker_ip is fully ready!\"
+    echo \"Worker \$worker_url is fully ready!\"
```
AryanAhadinia, line 296: Is this change and the changes above necessary?

src/swiss_ai_model_launch/assets/models.json
AryanAhadinia, line 1: Could you please also add the full command to `examples` directory?

AryanAhadinia (changes requested):
Thanks for your contribution! I would appreciate if you could also add the CLI command to the `examples` directory and then we could proceed with merging the PR.

AryanAhadinia:
Please note the ongoing change in #78 has changed model catalogue. `time` and `worker` entries are omitted.

src/swiss_ai_model_launch/assets/template.jinja
Alvorecer721: Yes — the health-check URL is the load-bearing fix; the variable rename and echo update are consequential.

The old code hardcoded `$WORKER_PORT` for every worker:
```bash
curl http://${worker_ip}:${WORKER_PORT}/health
```
That works for single-worker-per-node but hangs the router if two workers on the same host bind to different ports (e.g. `:5000` and `:5001`) — it polls `:5000` twice and never sees `:5001` come up. `worker_urls[*]` is already populated upstream with each worker's full `http://host:port`, so the correct form is:
```bash
curl ${worker_url}/health
```

The loop variable rename (`worker_ip` → `worker_url`) follows because it now holds a URL, not an IP — and the echo has to update to match, otherwise `$worker_ip` expands to empty inside the renamed loop and the log line breaks.

Happy to revert to `worker_ip` as the loop variable name if you prefer a smaller diff, but the name would then be misleading (holding a full URL, not an IP).

src/swiss_ai_model_launch/assets/models.json
Alvorecer721: Added in 7f2d0d1 — `examples/clariden/cli/qwen/Qwen3.6-27B-vllm.sh`. Follows the same pattern as the existing Qwen3-235B-vllm example, with DP=2/TP=2 on one GH200 node.

Alvorecer721: Done

AryanAhadinia (approved):

Issue #72, opened 2026-04-09
Load Test with/without Serving Stack


AryanAhadinia:
Run large scale loadtests while bypassing the serving stack to see the overhead of it in terms of latency/throughput. Need to run with/without `--disable-ocf`. Perhaps, is there a concurrency limit of the serving-api?

PR #98, opened 2026-04-24
Feature/DCGM Exporter

burakemirsezen:
Implements the first part of #72 

For this to work, vmagent needs to also scrape port 9400 and the csv file specifying which metrics are exported needs to be created. After creating all of this I was able to see the exports in grafana/prometheus.

Btw this has the same problem with vmagent, it currently only tracks the node that the job was launched on.

### DIFF
--- a/src/swiss_ai_model_launch/assets/template.jinja
+++ b/src/swiss_ai_model_launch/assets/template.jinja
@@ -28,6 +28,9 @@ SERVED_MODEL_NAME="{{ served_model_name }}"
 METRICS_REMOTE_WRITE_URL="{{ metrics_remote_write_url or '' }}"
 METRICS_AGENT_BIN="{{ metrics_agent_binary }}"
 DCGM_EXPORTER_BIN="{{ dcgm_exporter_binary }}"
+USE_DCGM_EXPORTER={{ "false" if disable_dcgm_exporter else "true" }}
+USE_METRICS={{ "false" if disable_metrics else "true" }}
+
 
 {% if telemetry_endpoint %}
 curl -sf -X POST "{{ telemetry_endpoint }}" \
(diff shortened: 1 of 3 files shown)


robmsmt:
Thanks @burakemirsezen 

>Btw this has the same problem with vmagent, it currently only tracks the node that the job was launched on.

This is only an issue for multi-node jobs correct? i.e. when the number of nodes>1. Can you suggest a fix?

burakemirsezen:
> Thanks @burakemirsezen
> 
> > Btw this has the same problem with vmagent, it currently only tracks the node that the job was launched on.
> 
> This is only an issue for multi-node jobs correct? i.e. when the number of nodes>1. Can you suggest a fix?

Yes this problem is only for multi-node jobs. This should be fixed if we just launch vmagent on every node of the job. I can look into that if you'd like.

burakemirsezen:
It's possible that this broke things. I know that the dcgm exports work for all nodes, but I couldn't try if the model works properly because the models don't load for some reason with or without this commit.

This now also needs to separate vmagent configs, one that scrapes the engine and one that doesn't.

robmsmt (approved):
Looks great, thanks for this. I think we should enable the user to disable this incase there are performance issues with recording DCGM? Can add it as a github issue for later if you want?

burakemirsezen:
> Looks great, thanks for this. I think we should enable the user to disable this incase there are performance issues with recording DCGM? Can add it as a github issue for later if you want?

Just dcgm or also vmagent in general?

I can add an env that disables them/it, is that enough?

Also, the dcgm exporter should be pretty lightweight and dcgm is already running so the resource usage should be pretty minimal.

Edit: Just added both.

AryanAhadinia (approved):

Issue #37, opened 2026-03-27
Add Comprehensive Telemetry


AryanAhadinia:
- Extend the data collection and collect data about the status of the launched models and the usage of that model.
- Develop analytics panels for knowledge discovery from the telemetry logs.

AryanAhadinia:
Also, make sure that every single launch is recorded in the telemetry DB.

PR #100, opened 2026-04-25
Refactor Job Submission Template

### LINKED ISSUE #27: `template.jinja` Refactor
template.jinja Likely needs a refactor. I see no reason it cannot be split into more manageable chunks. 

E.g. SGlang vs VLLM
simple single node vs multi

### LINKED ISSUE #134: updating the bootstrap ocf address --dev
currently to update the bootstrap address you have to edit https://github.com/swiss-ai/model-launch/blob/4d67436a37cefee88f69d2da453230bbd6c2cb2d/src/swiss_ai_model_launch/assets/template.jinja#L26

Would be nice to have shorthand --dev to use the other environment or be able to override for custom ocf in another datacenter/ip more easily.

AryanAhadinia:
## Summary

Refactor of the SLURM job submission pipeline. What started as Aryan's "drop Jinja, replace with static `script.sh` + env vars" has grown to:

1. **Terminology + topology cleanup** (replaces *worker* with *replica*, collapses `--slurm-nodes` / `--worker-port` knobs)
2. **Static `script.sh` deleted** — the SLURM script is now rendered end-to-end from Python (`launchers/framework.py`)
3. **Multi-file rank dispatch** — master.sh + per-shape `head.sh` / `follower.sh` / `router.sh`, self-extracted under `$HOME/.sml/job-${SLURM_JOB_ID}/` at job start
4. **Container-aware** — `--container-mounts="$RANKS_DIR:$RANKS_DIR"` binds the rank dir into pyxis containers (works with PR #124's `:ro,rprivate` mount tightening)
5. **`--output-script`** CLI flag for dry-run / debug
6. **Carries forward PR #124's vLLM Ray head OCF-subprocess fix** through the new renderer

See [ADR-0001](docs/adrs/0001-replica-is-the-orchestration-unit.md) for the *replica = orchestration unit* decision.

## Breaking-ish changes

| Before | After | Migration |
|---|---|---|
| `--slurm-workers` | `--slurm-replicas` | deprecation warning, old flag still accepted |
| `--slurm-nodes-per-worker` | `--slurm-nodes-per-replica` | same |
| `--slurm-nodes` | (gone — derived from `replicas × nodes_per_replica`) | warning, value ignored |
| `--worker-port`, `--replica-port` | (gone — hardcoded `FRAMEWORK_PORT=8080`) | warning, value ignored |
| `workers`, `nodes_per_worker` kwargs on `LaunchArgs`/`LaunchRequest`/`ModelCatalogEntry` | `topology.replicas`, `topology.nodes_per_replica` | pydantic `before` validator migrates with `DeprecationWarning` |
| MCP `launch_preconfigured_model(workers=…)` | `replicas=…` | **hard rename** (no MCP-side shim — single caller surface) |

## How the script is generated now

`framework.py` renders one self-contained `master.sh`. At job start, it:

1. `cat`-heredocs each rank shape to `$HOME/.sml/job-${SLURM_JOB_ID}/{head,follower,router}.sh` (single-quoted heredoc → bodies land verbatim, no escape soup)
2. Detects arch, maps nodes, discovers each replica's head IP
3. Per (replica, rank): `srun --container-mounts="$RANKS_DIR:$RANKS_DIR" --environment="..." bash $RANKS_DIR/head.sh "$replica_N_head_ip"`
4. Optional: launches sglang_router on `nodes[0]` (only when `replicas > 1 && --use-router`)
5. Optional: launches vmagent on the batch node (skippable via `--disable-metrics`)

The Ray-head-script-on-disk pattern from PR #124 is preserved verbatim inside the new `_render_vllm_head`.

## Testing

Six example scripts under `tests/unit/test_examples.py` are parsed through the production CLI argparse, run through the renderer, and validated via `bash -n` + `shellcheck`. Plus a 96-combo matrix in `test_framework.py` covering (framework × replicas × nodes_per_replica × use_router × disable_ocf × telemetry). **247 unit tests, all green.**

### End-to-end cluster validation

Ran on Clariden via firecrest + SSH/slurm paths. Every code path the refactor introduces is validated against real hardware:

| Test | Result |
|---|---|
| Singular sglang (Apertus 8B) | ✅ served, vmagent metrics flowing |
| Singular vLLM (Mistral-3-3B) | ✅ served |
| Multi-node sglang (Apertus 70B, TP=8) | ✅ served |
| Multi-node vLLM Ray (Qwen 235B, slurm path) | ❌ OOM during `torch.compile` — model-config issue, **NOT a refactor bug** (see #126 for a different Ray-bootstrap bug class fixed here) |
| Multi-node vLLM Ray (Qwen 235B, firecrest path) | ❌ same OOM |
| **Multi-replica + sglang_router** (2 × Apertus 8B) | ✅✅ router serving, both replicas health-checking, prefill batches running |
| `disable_ocf=True` | ✅ served, no OCF wrap |
| `pre_launch_cmds="pip install librosa audioread"` | ✅ Qwen3-ASR served |

## Test plan

- [x] `uv run pytest tests/unit/` — 247 pass
- [x] `make shellcheck` (via pre-commit hooks)
- [x] `uv run ruff check`, `ruff format`, `mypy` — clean
- [x] Live firecrest + slurm/SSH submissions on Clariden across 6 framework/topology shapes
- [ ] Re-verify the multi-replica router test once jobs aren't burning slurm time-limit credits


AryanAhadinia:
Regarding the separation of the script for vLLM and sglang, there is a trade-off. If we split the script in two for each framework, then we have a lot of duplication between them, and we cannot resolve duplication by extracting another shell script since the shell script is submitted through firecrest and only a single script can be submitted through firecrest. So, separating the shell scripts would result in more readable and shorter scripts, but more duplication.

My suggestion is to have it as a single script, but make them divided as much as possible for the sake of readability as the current state of the PR presents.

@robmsmt, what do you think?

robmsmt:
Great stuff, in a bit of a rush, feedback left on slack: https://swissai-initiative.slack.com/archives/C0ATZPVGCNM/p1777137480279369

AryanAhadinia:
Reply to @robmsmt review.

> - set -u + script reuse — if [ -n "$TELEMETRY_ENDPOINT" ] (line 9) dies under -u if the var isn't set.
  Fine when invoked through these launchers (always set to ""), but anyone running the script manually
  will be confused. Either default-init at the top (: "${TELEMETRY_ENDPOINT:=}") or document that it must
  be invoked via the launcher.

Fixed via default-initialization at the top of the script:

```shell
: "${TELEMETRY_ENDPOINT:=}"
: "${METRICS_REMOTE_WRITE_URL:=}"
: "${PRE_LAUNCH_CMDS:=}"
```

Also added METRICS_REMOTE_WRITE_URL and PRE_LAUNCH_CMDS which have the same exposure — both are optional vars guarded by [ -n "$VAR" ] checks that would also fire under set -u on manual invocation.

AryanAhadinia:
Reply to @robmsmt review.

> - No unit tests for the new pure functions — to_job_env(), to_sbatch_args(), render_sbatch_header() are trivially unit-testable (no slurm needed) and would catch field-rename regressions much faster than the integration suite. Worth adding before merge.

Added `tests/unit/test_launch_args.py` with 15 tests covering all three functions — no SLURM required:

- `to_job_env()`: key set completeness, value mapping, boolean encoding for `USE_ROUTER` / `USE_OCF`, `None` telemetry endpoint coerced to `""`, all values are strings.
- `to_sbatch_args()`: required flags present, `--nodes` auto-computed from `workers × nodes_per_worker`, `--reservation` included when set and absent when not.
- `render_sbatch_header()`: `#!/bin/bash` prefix, all `#SBATCH` directives, reservation present/absent.
- `resolve_model_path()`: registry-prefix construction and explicit-path override.

AryanAhadinia:
Reply to @robmsmt review.

> - Lost docstring — resolve_model_path in utils.py had a useful docstring explaining the registry-prefix
  behavior; it's been deleted with no obvious reason. Restore it.

Restored in `utils.py`. The docstring now explains both behaviors:

- If `model_path` is provided it is returned as-is, allowing callers to point at an arbitrary local copy.
- Otherwise the path is constructed by joining `registry` with `model` (e.g. `registry / "vendor/name"`), which is the standard layout used by the CSCS model registry.

AryanAhadinia:
Reply to @robmsmt review.

> - env={**os.environ, **[launch_args.to](http://launch_args.to/)_job_env()} in slurm_launcher.py:106 — this leaks the entire submitter shell environment into sbatch (and via slurm's default SBATCH_EXPORT=ALL, into the running job). Old behavior baked values into the script, so this is a behavior change. Recommend either passing only to_job_env() (plus PATH/HOME if needed) or explicitly setting --export=ALL,<vars> and inheriting nothing.

The behavior here is intentional and equivalent to submitting the job manually via `sbatch` from the terminal. SLURM's default `SBATCH_EXPORT=ALL` exports the full shell environment to the job in both cases, so there's no extra leakage. System variables like `$SCRATCH` and `$PATH` are also required by the job (the skybox plugin and tools like `hostname` depend on them). The `to_job_env()` values are merged on top, so model/framework config takes precedence over any conflicting keys from `os.environ`. Omitting `os.environ` is not possible — we verified that it causes the job to fail immediately due to missing `$SCRATCH` (skybox plugin) and `$PATH` (binary resolution).

AryanAhadinia:
@robmsmt, Shall we push it forward?

robmsmt (dismissed):

examples/clariden/cli/qwen/Qwen3-235B-A22B-Instruct-2507-vllm.sh
```diff
@@ -2,12 +2,11 @@
 sml advanced \
   --firecrest-system clariden \
   --partition normal \
-  --slurm-nodes 2 \
+  --slurm-nodes-per-replica 2 \
+  --slurm-time 02:00:00 \
```
AryanAhadinia: Why adding time? Time has been annoying recently since the default was 5 min. I think we should put default to 1 - 2 hours and omit time from all of the examples. We should make sure the default time is consistent across both advanced and interactive SML.

examples/clariden/cli/zai-org/GLM-4.6-sglang-router.sh
```diff
@@ -1,12 +1,11 @@
 #!/bin/bash
-# 2 workers x 4 nodes each. Requires latest sglang env. Experimental.
+# 2 replicas x 4 nodes each. Requires latest sglang env. Experimental.
```
AryanAhadinia: Stale comment I think.

pyproject.toml
```diff
@@ -93,4 +92,4 @@ show_missing = true
 fail_under = 0      # TODO
 
 [dependency-groups]
-dev = ["pytest-asyncio>=1.3.0", "pytest-xdist>=3.8.0"]
+dev = ["pre-commit>=4.6.0", "pytest-asyncio>=1.3.0", "pytest-xdist>=3.8.0"]
```
AryanAhadinia: This line is ultimately redundant since we have dev devs in line ~26 of this file. We should omit one of them.

pyproject.toml
AryanAhadinia: Maybe to jump the version to `26.5.0a1`?

pyproject.toml
```diff
@@ -13,7 +13,6 @@ requires-python = ">=3.10,<=3.14"
 dependencies = [
   "fastmcp>=3.2.3",
   "httpx>=0.28",
-  "jinja2>=3.0",
```
AryanAhadinia, line 16: Thanks god!

src/swiss_ai_model_launch/launchers/launch_request.py
```diff
@@ -11,21 +12,36 @@ class LaunchRequest(BaseModel):
...
+    nodes_per_replica: int
+    replicas: int
     time: str
     served_model_name: str | None = None
     framework_args: str | None = None
     pre_launch_cmds: str | None = None
     use_router: bool = False
     model_path: str | None = None
 
+    @model_validator(mode="before")
```
AryanAhadinia: I think this makes code complicated and we are not obligated to hold backward compatibility for now. Either to remove this block or mark this as _deprecated_ in a comment so that we note we should remove this soon.

docs/adrs/0001-replica-is-the-orchestration-unit.md
```diff
@@ -0,0 +1,59 @@
...
+All of these run **inside a single framework process** that exposes one HTTP endpoint.
+
+We could expose any of these as orchestration concepts, or none. We chose to expose none and call the framework process itself a *replica*.
+
+## Decision
+
+A **replica** is one independent inference engine instance — one framework process exposing one HTTP endpoint at port 8080. It is the unit of orchestration:
+
+- A SLURM job is `replicas × nodes_per_replica` nodes.
+- A replica spans a contiguous set of nodes (`nodes_per_replica`) and is wrapped in OCF on its head node.
```
AryanAhadinia, line 25: OpenTela?

docs/adrs/0001-replica-is-the-orchestration-unit.md
```diff
@@ -0,0 +1,59 @@
...
+- The router (sglang_router today) load-balances **across replicas**, not within them.
+- Internal sharding (TP/PP/DP/EP) is the user's concern, configured via free-form `framework_args`. SML does not infer or inject parallelism flags.
+
+## Consequences
+
+### Supported
+
+- Multi-replica fanout (`replicas >= 2`): N independent processes, each on its own node-set, router fans out across them. This is the canonical multi-tenant throughput pattern.
```
AryanAhadinia, line 33: Maybe to elaborate on which router, OpenTela router or SGLang/vLLM routers.

docs/adrs/0001-replica-is-the-orchestration-unit.md
```diff
@@ -0,0 +1,59 @@
...
+- Multi-node single-replica (`replicas=1, nodes_per_replica >= 2`): one framework process spread across multiple nodes via the framework's own distributed-init mechanism (sglang `--dist-init-addr`, vLLM Ray cluster).
+
+### Deliberately not supported
+
+- **Multiple framework processes on a single node.** A user might want to run 4×TP=1 sglang processes on a 4-GPU node and load-balance across them (via the router or OCF, both of which have the plumbing). To support this we would need:
```
AryanAhadinia, line 38: OpenTela

AryanAhadinia, line 38: 4xTP=1 is a little bit vague.

docs/adrs/0001-replica-is-the-orchestration-unit.md
```diff
@@ -0,0 +1,59 @@
...
+  - A new topology dimension (`processes_per_node` or similar).
+  - Per-process port allocation, undoing the hardcoded `FRAMEWORK_PORT=8080` (see ADR-0002 if/when written).
+  - Per-process OCF port allocation (8092/8093/43905 currently collide between OCF instances on a host).
+  - GPU pinning via `CUDA_VISIBLE_DEVICES`.
+  - User-level mental model: replicas-per-node vs replicas-per-job.
+  
+  This is real complexity bringing back exactly the kind of multi-source-of-truth port management we just collapsed. The optimisation it unlocks (DP fanout on a single host) is mostly already handled by sglang's internal scheduler with `--dp-size N`. We defer this until there's concrete demand.
```
AryanAhadinia, line 45: Out of context, worth to take a look to this for future iterations.

https://medium.com/@meshuggah22/stop-guessing-which-llms-fit-your-gpu-theres-a-tool-for-that-85a05e486cfd

docs/adrs/0001-replica-is-the-orchestration-unit.md
```diff
@@ -0,0 +1,59 @@
...
+
+- **`use_router=True` with `replicas=1`.** Validates as an error: there's nothing to load-balance across. (sglang's internal DP gives one HTTP endpoint regardless of how many DP workers it runs.)
+
+## Alternatives considered
+
+1. **Expose TP/PP/DP/EP as topology fields.** Rejected: parallelism choices are hardware/model-specific, easy to misconfigure into OOM, and frameworks evolve their own flags. SML staying out of this lets users follow framework docs without translation.
+
+2. **Allow `replicas=1` with `use_router=True` as a no-op or stable-URL pattern.** Rejected: router with one backend is pure overhead with no load-balancing value, and silently no-op'ing surprises users.
```
AryanAhadinia, line 53: maybe warning instead of error?

docs/development.md
```diff
@@ -82,6 +82,56 @@ The SML team can't take a "please add my model" request for every checkpoint tha
...
+
+1. **Telemetry** POST (omitted if no endpoint configured)
+2. **Arch detection** — sets `OCF_BIN`, `SP_NCCL_SO_PATH`, `metrics_agent_bin` per `aarch64` / `x86_64`
+3. **Node mapping** — `mapfile -t nodes < <(scontrol show hostnames ...)`
+4. **Self-extracting rank scripts** — single-quoted `cat`-heredocs that lay down `head.sh`, optionally `follower.sh`, optionally `router.sh` under `$HOME/.sml/job-${SLURM_JOB_ID}/`
+5. **Per-replica head IP discovery** — one `hostname -i` srun per replica
+6. **Per-rank `srun` calls** — one block per (replica, rank). Each binds the rank dir into the pyxis container via `--container-mounts="$RANKS_DIR:$RANKS_DIR"` and invokes `bash $RANKS_DIR/<role>.sh`
+7. **vmagent** (optional) — metrics scraper on the batch node
+8. **Router** (optional) — `sglang_router` on `nodes[0]` when `replicas > 1 && --use-router`
+9. **Footer** — connect/cancel hints, `wait`, "Script finished"
```
AryanAhadinia, line 101: This is a legacy from previous scripts. Maybe to omit these?

docs/development.md
```diff
@@ -82,6 +82,56 @@ The SML team can't take a "please add my model" request for every checkpoint tha
...
 
+## Modifying the SLURM submission script
+
+The SLURM script is **rendered from Python at submit time** — there is no static `script.sh` or `template.jinja` to edit. The renderer is in [`src/swiss_ai_model_launch/launchers/framework.py`](https://github.com/swiss-ai/model-launch/blob/main/src/swiss_ai_model_launch/launchers/framework.py).
+
+### What gets rendered
+
+A single `master.sh` (visible via `--output-script` — see [usage](usage-advanced.md#inspecting-what-would-be-submitted---output-script)) containing in order:
+
+1. **Telemetry** POST (omitted if no endpoint configured)
```
AryanAhadinia: > (omitted if no endpoint configured)

This is not the case I think. Also, we should omit this.

docs/development.md
```diff
@@ -82,6 +82,56 @@ The SML team can't take a "please add my model" request for every checkpoint tha
...
+sml advanced ... --output-script /tmp/before    # current behaviour
+# edit framework.py
+sml advanced ... --output-script /tmp/after     # new behaviour
+diff -r /tmp/before /tmp/after                  # per-file diff across master + ranks
+```
+
+For full coverage, the test matrix at `tests/unit/test_framework.py` renders 96 configurations (framework × replicas × nodes_per_replica × use_router × disable_ocf × telemetry) and runs `bash -n` + `shellcheck` against each. If your change leaves any of those broken, the test will catch it before submit time:
+
+```bash
+uv run pytest tests/unit/test_framework.py -q
```
AryanAhadinia, line 130: Maybe add the command for running all the tests.

docs/usage-advanced.md
```diff
@@ -18,37 +18,81 @@ For the guided flow with a curated catalog, use [`sml`](usage-sml.md).
...
+| `--use-router`              |                         | Load-balance across replicas (needs `replicas > 1`)               |
 | `--router-args`             |                         | Arguments forwarded to the router                                 |
 | `--disable-ocf`             |                         | Disable OCF wrapper                                               |
+| `--disable-metrics`         |                         | Disable vmagent metrics push                                      |
+| `--disable-dcgm-exporter`   |                         | Disable DCGM GPU metrics exporter                                 |
 | `--pre-launch-cmds`         |                         | Shell commands to run before the framework starts                 |
+| `--output-script DIR`       |                         | Render master.sh + rank scripts into DIR and exit (no submit)     |
+
+> Total nodes is automatically `--slurm-replicas × --slurm-nodes-per-replica`; there is no separate `--slurm-nodes` flag. The framework HTTP port is hardcoded to **8080** across every job — no `--worker-port`/`--replica-port` knob.
```
AryanAhadinia: I think this is good to mention, but not here. It is kind of aligned with our mentality because we have just omitted that from here, but if we re-think from scratch, it is not that much related to here.

docs/usage-advanced.md
```diff
@@ -18,37 +18,81 @@ For the guided flow with a curated catalog, use [`sml`](usage-sml.md).
...
   --partition normal \
-  --slurm-replicas 1 \
-  --slurm-nodes-per-replica 1 \
   --serving-framework sglang \
   --slurm-environment src/swiss_ai_model_launch/assets/envs/sglang.toml \
   --framework-args "--model-path /capstor/store/cscs/swissai/infra01/hf_models/models/swiss-ai/Apertus-8B-Instruct-2509 \
     --served-model-name swiss-ai/Apertus-8B-Instruct-2509-$(whoami) \
     --host 0.0.0.0 \
-    --port 8080"
+    --enable-metrics"
```
AryanAhadinia, line 48: Maybe having the examples as minimal as possible is the best.

docs/usage-advanced.md
```diff
@@ -18,37 +18,81 @@ For the guided flow with a curated catalog, use [`sml`](usage-sml.md).
...
 sml advanced \
   --firecrest-system clariden \
   --partition normal \
-  --slurm-replicas 1 \
-  --slurm-nodes-per-replica 1 \
   --serving-framework sglang \
   --slurm-environment src/swiss_ai_model_launch/assets/envs/sglang.toml \
   --framework-args "--model-path /capstor/store/cscs/swissai/infra01/hf_models/models/swiss-ai/Apertus-8B-Instruct-2509 \
     --served-model-name swiss-ai/Apertus-8B-Instruct-2509-$(whoami) \
     --host 0.0.0.0 \
```
AryanAhadinia, line 47: We can also omit host I think since it is everywhere.

docs/usage-advanced.md
```diff
@@ -18,37 +18,81 @@ For the guided flow with a curated catalog, use [`sml`](usage-sml.md).
...
+
+```bash
+sml advanced \
+  --firecrest-system clariden \
+  --partition normal \
+  --serving-framework sglang \
+  --slurm-environment src/swiss_ai_model_launch/assets/envs/sglang.toml \
+  --framework-args "--model-path /capstor/.../Apertus-8B-Instruct-2509 \
+    --served-model-name swiss-ai/Apertus-8B-Instruct-2509-$(whoami) \
+    --host 0.0.0.0 --enable-metrics" \
```
AryanAhadinia, line 67: next line? (or preferably omitted to keep the example minimal)

docs/usage-advanced.md
```diff
@@ -18,37 +18,81 @@ For the guided flow with a curated catalog, use [`sml`](usage-sml.md).
...
+
+`--output-script DIR` writes the rendered submission scripts into the given directory and exits without submitting:
+
+```bash
+sml advanced \
+  --firecrest-system clariden \
+  --partition normal \
+  --serving-framework sglang \
+  --slurm-environment src/swiss_ai_model_launch/assets/envs/sglang.toml \
+  --framework-args "--model-path /capstor/.../Apertus-8B-Instruct-2509 \
```
AryanAhadinia, line 65: To avoid quote collisions, I was thinking maybe we can change the framework args and pass them just as ordinary args.

```
sml advanced \
  --firecrest-system clariden \
  --partition normal \
  --serving-framework sglang \
  --slurm-environment src/swiss_ai_model_launch/assets/envs/sglang.toml \
  --model-path /capstor/.../Apertus-8B-Instruct-2509 \
  --served-model-name swiss-ai/Apertus-8B-Instruct-2509-$(whoami) \
  --host 0.0.0.0 \
  --enable-metrics
```

`sml` knows what args it has, then it can pass the rest to the framework.

docs/usage-advanced.md
```diff
@@ -18,37 +18,81 @@ For the guided flow with a curated catalog, use [`sml`](usage-sml.md).
...
+diff /tmp/debug/head.sh /tmp/older-debug/head.sh   # compare runs
+```
+
+Useful for:
+
+- **Debugging a launch failure:** see exactly what `--framework-args` your invocation translated to (the `--port 8080` auto-injection, etc.), which `srun` calls would run, and what each rank shape does.
+- **Reviewing changes during SML development:** render against a known invocation before and after a code change, diff the rank scripts.
+- **Starting point for a hand-tuned job:** edit any of the rank scripts, then `sbatch master.sh` directly.
+
+After a real (non-`--output-script`) submission, the same rank scripts also land on disk at `~/.sml/job-${SLURM_JOB_ID}/` for post-mortem inspection.
```
AryanAhadinia, line 94: Maybe we should consider putting logs, uploaded toml file (for FirecREST mode), and the script all together.

docs/usage-advanced.md
```diff
@@ -18,37 +18,81 @@ For the guided flow with a curated catalog, use [`sml`](usage-sml.md).
...
+
+> **`master.sh` is self-contained.** It carries the rank scripts as embedded `cat`-heredocs and re-extracts them under `~/.sml/job-${SLURM_JOB_ID}/` at job start — that's why `sbatch master.sh` works regardless of where the directory lives. The standalone `head.sh` / `follower.sh` / `router.sh` files in the output dir are byte-equal duplicates for inspection; the heredoc bodies inside `master.sh` are what actually run.
```
AryanAhadinia: Maybe elaborate a bit more on self-containment and how it will help us.

docs/usage-advanced.md
```diff
@@ -18,37 +18,81 @@ For the guided flow with a curated catalog, use [`sml`](usage-sml.md).
...
+
 ## When to disable OCF
 
 > "OCF" and "OpenTela" refer to the same thing — `OCF` is the on-disk binary name from the [OpenTela project](https://github.com/swiss-ai/opentela). The flag is `--disable-ocf` for historical reasons.
```
AryanAhadinia, line 100: There is a PR from Xiazhe for renaming OCF to OpenTela in the whole repo. Let's just use OTela everywhere from now on.

src/swiss_ai_model_launch/launchers/model_catalog_entry.py
```diff
@@ -9,7 +10,19 @@ class ModelCatalogEntry(BaseModel):
     model: str
     framework: Literal["sglang", "vllm"]
     environment: str | None = None
-    nodes_per_worker: int = 1
+    nodes_per_replica: int = 1
     framework_args: str | None = None
     pre_launch_cmds: str | None = None
     model_path: str | None = None
+
+    @model_validator(mode="before")
```
AryanAhadinia: https://github.com/swiss-ai/model-launch/pull/100/changes#r3222421182

examples/clariden/cli/swiss-ai/1p5-experiments/Apertus-1.5-8B-cooldown-1pct-vision-sft.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+sml advanced \
+  --firecrest-system clariden \
+  --partition normal \
+  --slurm-reservation SD-69241-apertus-1-5 \
```
AryanAhadinia: Warning: Reservation

src/swiss_ai_model_launch/cli/main.py
```diff
@@ -32,6 +33,16 @@
 _OptionsFactory = Callable[[], Awaitable[OptionsDict]] | Callable[[GetValueFn], Awaitable[OptionsDict]] | None
 
 
+def _resolve_legacy(
```
AryanAhadinia: Repeated: no need for BWC for now I think.

src/swiss_ai_model_launch/cli/main.py
```diff
@@ -179,26 +190,43 @@ def _build_parser() -> argparse.ArgumentParser:
         metavar="ARGS",
         help="Arguments forwarded to the inference framework.",
     )
+    advanced_parser.add_argument(
+        "--slurm-replicas",
```
AryanAhadinia, line 187: minor: the num replica and node per replica is in an abstraction level higher than slurm. Maybe we can just use `replicas` and `nodes-per-replica`.

src/swiss_ai_model_launch/cli/main.py
```diff
@@ -545,29 +601,100 @@ async def _run_advanced(args: argparse.Namespace) -> None:
...
+        account=launcher.account,
+        partition=launcher.partition,
         telemetry_endpoint=config.get_value("telemetry_endpoint"),
     )
 
+    if args.output_script:
+        from pathlib import Path
+
+        from swiss_ai_model_launch.launchers.framework import render_master, render_rank_scripts
+        from swiss_ai_model_launch.launchers.utils import render_sbatch_header
```
AryanAhadinia: Suggest to use ordinary import instead of lazy import.

src/swiss_ai_model_launch/cli/main.py
```diff
@@ -545,29 +601,100 @@ async def _run_advanced(args: argparse.Namespace) -> None:
...
+        # `sbatch <dir>/master.sh` is also a valid way to submit manually.
+        out_dir = Path(args.output_script).expanduser().resolve()
+        out_dir.mkdir(parents=True, exist_ok=True)
+
+        master_path = out_dir / "master.sh"
+        master_path.write_text(render_sbatch_header(launch_args) + render_master(launch_args))
+        master_path.chmod(0o755)
+
+        written = ["master.sh"]
+        for filename, content in render_rank_scripts(launch_args).items():
```
AryanAhadinia, line 646: It is a bit concerning for me. Why we are having multiple files? Wouldn't it be problematic for multi-node job submission using FirecREST?

src/swiss_ai_model_launch/launchers/framework.py
```diff
@@ -0,0 +1,600 @@
...
+no nested quoting concerns. The master invokes them via ``bash <path> args...``
+with positional args for the runtime-determined values (``$replica_N_head_ip``,
+``$node_rank``).
+
+This shape keeps the worst pre-existing pain — bash text embedded inside a
+``bash -c "..."`` arg whose outer-shell processing forces manual escape
+juggling — entirely out of the picture.
+"""
+
+from __future__ import annotations
```
AryanAhadinia, line 1: Don't like importing like that. I think Claude will be able to omit that.

src/swiss_ai_model_launch/launchers/framework.py
```diff
@@ -0,0 +1,600 @@
...
+
+Each rank script is a normal idiomatic bash file: shellcheckable on its own,
+no nested quoting concerns. The master invokes them via ``bash <path> args...``
+with positional args for the runtime-determined values (``$replica_N_head_ip``,
+``$node_rank``).
+
+This shape keeps the worst pre-existing pain — bash text embedded inside a
+``bash -c "..."`` arg whose outer-shell processing forces manual escape
+juggling — entirely out of the picture.
+"""
```
AryanAhadinia: Maybe not to have docs at the header of the pages.

src/swiss_ai_model_launch/launchers/framework.py
AryanAhadinia, line 1: The file is so large, maybe we can split it into a package with subfiles like `vllm.py`, `sglang.py`, etc.

src/swiss_ai_model_launch/launchers/framework.py
```diff
@@ -0,0 +1,600 @@
...
+    LaunchArgs,
+)
+
+SGLANG_ROUTER_PORT = 30000
+OCF_BOOTSTRAP_ADDR = "/ip4/148.187.108.178/tcp/43905/p2p/QmbUKJkCfotDzbFE5uoTsXD4GRyPHjzZC1f2yAGLoeBMn9"
+RAY_PORT = 6379
+NUM_GPUS_PER_NODE = 4
+SGLANG_DIST_INIT_PORT = 5757
+
+_VMAGENT_SCRAPE_CONFIG = "/capstor/store/cscs/swissai/infra01/ocf-share/vmagent-scrape.yaml"
```
AryanAhadinia, line 22: Why this global var is private while others are marked as public?

src/swiss_ai_model_launch/launchers/framework.py
```diff
@@ -0,0 +1,600 @@
...
+
+from typing import ClassVar
+
+from swiss_ai_model_launch.launchers.launch_args import (
+    FRAMEWORK_PORT,
+    LaunchArgs,
+)
+
+SGLANG_ROUTER_PORT = 30000
+OCF_BOOTSTRAP_ADDR = "/ip4/148.187.108.178/tcp/43905/p2p/QmbUKJkCfotDzbFE5uoTsXD4GRyPHjzZC1f2yAGLoeBMn9"
```
AryanAhadinia, line 16: I was thinking that maybe we should expose this as env var so we can easily overrwrite it for dev purposes.

src/swiss_ai_model_launch/launchers/framework.py
```diff
@@ -0,0 +1,600 @@
...
+    name = "vllm"
+    entrypoint = "python3 -m vllm.entrypoints.openai.api_server"
+    env_exports = [
+        "export RAY_CGRAPH_get_timeout=1800",
+        'export no_proxy="0.0.0.0,$no_proxy"',
+        'export NO_PROXY="0.0.0.0,$NO_PROXY"',
+    ]
+
+
+_FRAMEWORKS: dict[str, type[Framework]] = {"sglang": Sglang, "vllm": Vllm}
```
AryanAhadinia, line 55: I think it is not necessary and not that much clean.

src/swiss_ai_model_launch/launchers/framework.py
```diff
@@ -0,0 +1,600 @@
...
+    pre = _shebang_and_setup(framework, launch_args.pre_launch_cmds)
+    use_ocf = not launch_args.disable_ocf
+
+    if npr == 1:
+        # Singular: one rank per replica, the head IS the only rank.
+        # ``$1`` is replica_head_ip (passed by master, unused here but kept
+        # for signature symmetry with the multi-node case).
+        body_args = '# shellcheck disable=SC2034\nreplica_head_ip="$1"\n'
+        cmd = f"{framework.entrypoint} {args}"
+    else:
```
AryanAhadinia, line 154: We can use multiline strings + dedent.

src/swiss_ai_model_launch/launchers/framework.py
```diff
@@ -0,0 +1,600 @@
...
+        '"interactive": false, '
+        f'"serving_framework": "{launch_args.framework}", '
+        f'"framework_args": "{fa}", '
+        f'"pre_launch_cmds": "{launch_args.pre_launch_cmds}", '
+        f'"model_name": "{launch_args.served_model_name}", '
+        f'"replicas": {topology.replicas}, '
+        f'"nodes_per_replica": {topology.nodes_per_replica}, '
+        f'"framework_port": {FRAMEWORK_PORT}, '
+        f'"use_router": {use_router}, '
+        f'"router_environment": "{launch_args.environment}", '
```
AryanAhadinia, line 336: We should think about quote safety of the telemetry payload.

tests/unit/test_launch_args.py
```diff
@@ -0,0 +1,233 @@
...
+def test_to_job_env_telemetry_endpoint_none():
+    assert _make_args(telemetry_endpoint=None).to_job_env()["TELEMETRY_ENDPOINT"] == ""
+
+
+def test_to_job_env_all_values_are_strings():
+    for k, v in _make_args().to_job_env().items():
+        assert isinstance(v, str), f"{k} is not a string"
+
+
+# ── to_sbatch_args ───────────────────────────────────────────────────────────
```
AryanAhadinia: I don't like this comments. Either to split them in files or just simply omit these lines.

src/swiss_ai_model_launch/launchers/launch_args.py
```diff
@@ -1,4 +1,35 @@
...
+# The framework's HTTP server port is hardcoded across the system: it's
+# auto-injected as ``--port`` into framework_args, used as OCF's
+# ``--service.port``, and embedded in the router's worker URLs. Exposing
+# it as a knob just creates ways for the three to drift.
+FRAMEWORK_PORT = 8080
+
+_PORT_FLAG_RE = re.compile(r"(?:^|\s)--port(?:[\s=])")
+
+
+class Topology(BaseModel):
```
AryanAhadinia: Suggesting having a separate file for each class (Java-like)

AryanAhadinia (commented):
My main concern is about job submission with FirecREST. Nevertheless, I am working on automated tests. We will merge it with automated tests.

src/swiss_ai_model_launch/launchers/framework.py
robmsmt: suggestion?

tests/unit/test_launch_args.py
robmsmt: agree

src/swiss_ai_model_launch/launchers/launch_request.py
robmsmt: Agreed, good point

docs/adrs/0001-replica-is-the-orchestration-unit.md
robmsmt: Separate PR / issue

robmsmt: Disagree, I think the text makes this clear what it is?

docs/adrs/0001-replica-is-the-orchestration-unit.md
robmsmt: install it as an API, but i think this is better in issue or another PR

docs/adrs/0001-replica-is-the-orchestration-unit.md
robmsmt: error is fine? What is the advantage of having use_router with replica=1

docs/usage-advanced.md
robmsmt: agree

docs/usage-advanced.md
robmsmt: goal isn't to make the smallest possible config. I don't think we can remove it?

docs/usage-advanced.md
robmsmt: disagree on minimalism

docs/usage-advanced.md
robmsmt: Not a bad idea but we should save it for another release / issue

docs/usage-advanced.md
robmsmt: Agreed good idea.. it means we could relaunch it exactly? Assuming it wasn't modified a lot.

Can you make an issue?

docs/usage-advanced.md
robmsmt: I think this is now outdated. We should change it

docs/usage-advanced.md
robmsmt: Agree but in separate PR

examples/clariden/cli/qwen/Qwen3-235B-A22B-Instruct-2507-vllm.sh
AryanAhadinia: Deleted times from all of the cli examples (as well as reservation and `--dev` flag).

pyproject.toml
AryanAhadinia: Fixed the pyproject.toml structure and the corresponding references.

pyproject.toml
AryanAhadinia: Done

docs/adrs/0001-replica-is-the-orchestration-unit.md
AryanAhadinia: For dev purposes and benchmarking it may be useful, but let's leave it now as it is.

src/swiss_ai_model_launch/launchers/framework.py
AryanAhadinia: Omit telemetry step in init step as discussed in call with @robmsmt.

examples/clariden/cli/zai-org/GLM-4.6-sglang-router.sh
AryanAhadinia: Deleted all the stale comments on the examples scripts.

src/swiss_ai_model_launch/launchers/framework.py
AryanAhadinia: done with `--dev`

src/swiss_ai_model_launch/launchers/framework.py
AryanAhadinia: moved to #37.

PR #129, opened 2026-05-13
adjusted the scripts from apertus pt+sft+align

robmsmt:


robmsmt:
Merged into the AryanAhadinia/JobTemplateRefactor branch (commit c3aab6d), which is about to land. Closing here to avoid duplicate merge.

Issue #126, opened 2026-05-07
vLLM Launch Crashes for Multi-node Models


AryanAhadinia:
## Bug Description

**What happened?**

Error: Got unexpected extra arguments (& RAY_HEAD_PID= echo 'Waiting for all Ray worker nodes to connect...' EXPECTED_GPUS=8 while true; do AVAILABLE_GPUS=0 echo Available)

**What did you expect to happen?**

Normal launch

## Reproduction Steps

**Does this bug occur consistently or only occasionally?**

- [X] Consistently (happens every time)
- [ ] Occasionally (intermittent / hard to reproduce)

1. Run `examples/clariden/cli/qwen/Qwen3-235B-A22B-Instruct-2507-vllm.sh`

**Minimal `sml` command to reproduce (if applicable):**

```bash
examples/clariden/cli/qwen/Qwen3-235B-A22B-Instruct-2507-vllm.sh
```

## Environment

**SML version** (run `sml --version`): latest at the time.

**Framework** (sglang / vllm / other): vLLM

**Model** (e.g. `Qwen/Qwen3-235B-A22B-Instruct-2507`, or N/A): `Qwen3-235B-A22B-Instruct-2507`

**Cluster / Partition**: Clariden

## Logs and Error Output

Paste ALL and ANY relevant error messages or logs below.

```text
Usage: ray start [OPTIONS]
Try 'ray start --help' for help.

Error: Got unexpected extra arguments (& RAY_HEAD_PID= echo 'Waiting for all Ray worker nodes to connect...' EXPECTED_GPUS=8 while true; do AVAILABLE_GPUS=0 echo Available)
[2026-05-07 19:45:38,747 W 11614 11614] rpc_client.h:153: Failed to connect to GCS at address 172.28.16.184:6379 within 5 seconds.
2026-05-07T19:46:03.016+0200    ERROR   protocol/clock.go:70    Health check failed
srun: error: nid005422: task 0: Exited with exit code 1
srun: Terminating StepId=2060384.2
[2026-05-07 19:46:08,748 W 11614 11614] gcs_client.cc:205: Failed to get cluster ID from GCS server: TimedOut: Timed out while waiting for GCS to become available.
[2026-05-07 19:46:14,753 W 11614 11614] rpc_client.h:153: Failed to connect to GCS at address 172.28.16.184:6379 within 5 seconds.
```

## Checklist

- [X] I have confirmed the reproducibility above.

## Additional Context

[...]

Issue #118, opened 2026-05-04
Docker Golden Images


AryanAhadinia:
- Build Docker images on top of CSCS golden images.
- Benchmark the performance of the LLMs and ensure they are properly running on libfabric.
- Comprehend Docker CI pipeline. Proper versioning and arch annotation.

PR #135, opened 2026-05-20
Add Qwen3-TTS VoiceDesign vLLM-Omni launcher

yfchoco208:
Adds `examples/clariden/cli/qwen/Qwen3-TTS-12Hz-1.7B-VoiceDesign-vllm-omni.sh`, single-node launcher for [`Qwen/Qwen3-TTS-12Hz-1.7B-VoiceDesign`](https://github.com/QwenLM/Qwen3-TTS), serving text-to-speech via vLLM-Omni on Clariden GH200.

Adds `images/vllm_qwen3_tts_cuda13/Dockerfile` and `src/swiss_ai_model_launch/assets/envs/vllm_qwen3_tts_cuda13.toml`, a CUDA13 vLLM-Omni TTS environment with `vllm==0.20.2`, `vllm-omni==0.20.0`, `transformers==5.8.0`, and audio dependencies such as FFmpeg, libsndfile, and soundfile.

Adds `Qwen/Qwen3-TTS-12Hz-1.7B-VoiceDesign` to `src/swiss_ai_model_launch/assets/models.json`, an interactive SML catalog entry using vLLM-Omni with `--max-model-len 8192` and `--gpu-memory-utilization 0.40`. VoiceDesign was tested with `task_type=VoiceDesign` and text instructions rather than preset CustomVoice speakers.

Also adds `vllm-omni` as a supported framework where required, matching the existing vLLM-Omni serving pattern.

Validated from a clean checkout:
- `sml advanced` launch works
- interactive `sml` catalog launch works

### DIFF
--- a/examples/clariden/cli/funaudiollm/Fun-CosyVoice3-0.5B-2512-vllm-omni.sh
+++ b/examples/clariden/cli/funaudiollm/Fun-CosyVoice3-0.5B-2512-vllm-omni.sh
@@ -1,0 +1,36 @@
+#!/bin/bash
+# Launch Fun-CosyVoice3 0.5B with vLLM on one Clariden GH200 node.
+#
+# Model: FunAudioLLM/Fun-CosyVoice3-0.5B-2512
+#
+# Note:
+#   CosyVoice3 requires ref_audio/ref_text for /v1/audio/speech.
+#   If using local file:// reference audio, set:
+#     export COSYVOICE_ALLOWED_LOCAL_MEDIA_PATH=/path/to/reference/audio/root
+
+EXTRA_MEDIA_ARGS=""
+
+if [ -n "${COSYVOICE_ALLOWED_LOCAL_MEDIA_PATH:-}" ]; then
+  EXTRA_MEDIA_ARGS="--allowed-local-media-path ${COSYVOICE_ALLOWED_LOCAL_MEDIA_PATH}"
+fi
+
+sml advanced \
+  --firecrest-system clariden \
+  --partition normal \
+  --slurm-nodes 1 \
+  --slurm-time 6:00:00 \
+  --serving-framework vllm \
+  --served-model-name FunAudioLLM/Fun-CosyVoice3-0.5B-2512 \
+  --worker-port 8080 \
+  --slurm-environment src/swiss_ai_model_launch/assets/envs/vllm_cuda13_v2.toml \
+  --framework-args "FunAudioLLM/Fun-CosyVoice3-0.5B-2512 \
+    --host 0.0.0.0 \
+    --port 8080 \
+    --deploy-config /opt/conda/lib/python3.12/site-packages/vllm_omni/deploy/cosyvoice3.yaml \
+    --omni \
+    --trust-remote-code \
+    --enforce-eager \
+    --max-model-len 8192 \
+    --gpu-memory-utilization 0.40 \
+    ${EXTRA_MEDIA_ARGS}" \
+  --no-tui
--- a/examples/clariden/cli/qwen/Qwen3-TTS-12Hz-1.7B-VoiceDesign-vllm-omni.sh
+++ b/examples/clariden/cli/qwen/Qwen3-TTS-12Hz-1.7B-VoiceDesign-vllm-omni.sh
@@ -1,0 +1,24 @@
+#!/bin/bash
+# Launch Qwen3-TTS 12Hz 1.7B VoiceDesign with vLLM on one Clariden GH200 node.
+#
+# Model: Qwen/Qwen3-TTS-12Hz-1.7B-VoiceDesign
+
+sml advanced \
+  --firecrest-system clariden \
+  --partition normal \
+  --slurm-nodes 1 \
+  --slurm-time 6:00:00 \
+  --serving-framework vllm \
+  --served-model-name Qwen/Qwen3-TTS-12Hz-1.7B-VoiceDesign \
+  --worker-port 8080 \
+  --slurm-environment src/swiss_ai_model_launch/assets/envs/vllm_cuda13_v2.toml \
+  --framework-args "Qwen/Qwen3-TTS-12Hz-1.7B-VoiceDesign \
+    --host 0.0.0.0 \
+    --port 8080 \
+    --deploy-config /opt/conda/lib/python3.12/site-packages/vllm_omni/deploy/qwen3_tts.yaml \
+    --omni \
+    --trust-remote-code \
+    --enforce-eager \
+    --max-model-len 8192 \
+    --gpu-memory-utilization 0.40" \
+  --no-tui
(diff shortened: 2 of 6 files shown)


src/swiss_ai_model_launch/assets/template.jinja
```diff
@@ -83,6 +83,10 @@ case "$FRAMEWORK" in
         FRAMEWORK_ENV_SETUP="export RAY_CGRAPH_get_timeout=1800; export no_proxy=\"0.0.0.0,\$no_proxy\"; export NO_PROXY=\"0.0.0.0,\$NO_PROXY\""
         FRAMEWORK_LAUNCH="python3 -m vllm.entrypoints.openai.api_server"
         ;;
+    vllm-omni)
+        FRAMEWORK_ENV_SETUP="export RAY_CGRAPH_get_timeout=1800; export no_proxy=\"0.0.0.0,\$no_proxy\"; export NO_PROXY=\"0.0.0.0,\$NO_PROXY\""
+        FRAMEWORK_LAUNCH="vllm serve"
+        ;;
```
AryanAhadinia: This line indeed seems redundant to me as it is identical to the vLLM case. We may change `python3 -m vllm.entrypoints.openai.api_server` with `vllm serve` as they are identical since the former one is deprecated. Nevertheless, please note that we have massively refactored the codebase in #100 and the `template.jinja` file is now completely removed. Instead, we are now rendering the job script during the runtime in `framework.py`.

src/swiss_ai_model_launch/launchers/launch_request.py
```diff
@@ -9,7 +9,7 @@ class LaunchRequest(BaseModel):
     """A fully-specified launch request — catalogue fields plus user-supplied runtime parameters."""
 
     model: str
-    framework: Literal["sglang", "vllm"]
+    framework: Literal["sglang", "vllm", "vllm-omni"]
```
AryanAhadinia: Adding vLLM OMNI beside vLLM as a new framework should be well justified. In the long vision we have, we would like to have to golden base images for vLLM and SGL (ref: #118). As a result, I would suggest to drop `vllm-omni` as a new framework for now and just use (`--environment`/`--slurm-environment`) to specify which `toml` file you want to use.

images/vllm_qwen3_tts_cuda13/Dockerfile
AryanAhadinia, line 1: Isn't it possible to patch the current vLLM image?

AryanAhadinia (changes requested):
Thanks a lot for your contribution! We would love to merge your PR after addressing the listed comments! Keep doing the great job!

Please also note that your PR has now conflicts that should be solved prior to merge.

src/swiss_ai_model_launch/launchers/launch_request.py
yfchoco208: Thank you for clarifying, I will remove vllm-omni as new framework and stick to using the original vllm

images/vllm_qwen3_tts_cuda13/Dockerfile
yfchoco208: Just to clarify what you mean by “patch the current vLLM image”?

I'm not sure if you meant one of the following:

1. Use existing Docker vLLM CUDA13 base image if it exists and make `vllm_qwen3_tts_cuda13` (derived image) that only adds vllm-omni and audio dependencies.
2. Modify the current `vllm_cuda13` Dockerfile (image) itself to include vllm-omni and audio dependencies.

AryanAhadinia: The second one. In general, we are working toward keeping the number of images and environment as minimal as possible. So, adding a new image and environment only for a small class of models is not that much aligned with our long-term goals.

yfchoco208: I worked on patching the existing `images/vllm_cuda13/Dockerfile`.

* Removed the idea of adding `vllm-omni` as a new SML framework.
* Switched the Qwen3-TTS entry to use the existing `framework: vllm` and made the launch path to use the existing vllm environment pattern.

I found a compatibility issue I found is that adding `vllm-omni==0.20.0` on top of the current vllm cuda13 caused import failure because the current image has `vllm 0.21.1rc...`, when `vllm-omni==0.20.0` expects the vLLM 0.20 API layout, so i validated with the following combination:

* `vllm==0.20.2`
* `vllm-omni==0.20.0`
* `transformers==5.8.0`

I tested the patched `vllm_cuda13` image using a temporary `.sqsh` so I did not overwrite the shared image, and tested that Qwen3-TTS VoiceDesign starts and `/v1/audio/speech` generates WAV output, and normal vLLM text model (`swiss-ai/Apertus-8B-Instruct-2509`) also starts with `--enforce-eager`, and `/v1/chat/completions` returns a valid response.

Since the patch seems to be working with Qwen3-TTS and other models that use the docker image, can I proceed with patching the existing `vllm_cuda13` Dockerfile? 

But I just want to verify if it is okay to pin the shared vllm cuda13 image to `vllm==0.20.2` for compatibility with `vllm-omni==0.20.0` or not.

AryanAhadinia: Thanks a lot! What I would like to suggest is to create a new Dockerfile `vllm_cuda13_v2` and drop your patched Dockerfile there. Since the changes are now minimal, it would be hopefully easy for us to replace the original image with yours.

There are ongoing efforts in fixing some bugs that we have with vLLM (#126) and building golden docker images (#118 and #93). The complete replacement of the current vLLM image with yours will take place after the resolution of the aforementioned issues and PRs.

Please note that the CI pipeline should automatically build the Docker image for your and place it beside the other docker images.

Again, thanks a lot for your great job!

yfchoco208: Hi, I've pushed the updated version, but the two remaining CI failures seem likely to be related to repository CI configuration/key.

For `Docker Build vllm_cuda13_v2`, the job seem to fail during FirecREST initialization before the image build and gives error message:

`requests.exceptions.MissingSchema: Invalid URL '': No scheme supplied. Perhaps you meant https://?`

I manually built and tested `vllm_cuda13_v2` on Clariden using a temporary sqsh and toml, but the official GitHub image build seems to require repository FirecREST secrets to build `vllm_cuda13_v2.sqsh`.

For `SonarCloud / analyze`, the message also suggests a missing token/project permission issue:

`Warning: Running this GitHub Action without SONAR_TOKEN is not recommended`

`Project not found. Please check the 'sonar.projectKey' and 'sonar.organization' properties, the 'SONAR_TOKEN' environment variable, or contact the project administrator to check the permissions of the user the token belongs to`

I would greatly appreciate if you can help me solve this issue, whether if this is an issue about the dockerfile or issue about missing secret/key, Thank you!

AryanAhadinia:
The CI pipeline is failing. @yfchoco208, could you please rebase your branch to latest main?
```bash
git fetch
git rebase origin/main
git push --force
```

AryanAhadinia:
Replaced by #183.

PR #138, opened 2026-05-21
add deepseek v4 attempt

robmsmt:
## Summary

Adds two `sml advanced` launch examples for DeepSeek-V4-Pro on Clariden, plus the **vLLM image needed to serve them**:

- `examples/clariden/cli/deepseek-ai/DeepSeek-V4-Pro-vllm.sh` — post-trained V4-Pro, 4 nodes × TP16
- `examples/clariden/cli/deepseek-ai/DeepSeek-V4-Pro-Base-vllm.sh` — V4-Pro-Base (1.6T), 8 nodes × TP32
- `images/vllm_deepseek_v4/Dockerfile` — vLLM serving image with DeepSeek-V4 support (auto-built by CI)
- `src/swiss_ai_model_launch/assets/envs/vllm_deepseek_v4.toml` — env pointing at the built image

## Why a new image (the original sglang attempt failed)

The first attempt used `--serving-framework sglang` with the shared `sglang.toml`. Both jobs failed at sglang startup:

```
ValueError: The checkpoint you are trying to load has model type `deepseek_v4`
but Transformers does not recognize this architecture.
KeyError: 'deepseek_v4'
```

Root cause, verified in-container on Clariden:

| Image | sglang/vllm | transformers | `deepseek_v4`? |
|-------|-------------|--------------|----------------|
| `sglang_cuda13.sqsh` | sglang 0.5.10.post1 (no `DeepseekV4ForCausalLM`) | 5.3.0 (no `deepseek_v4`) | ❌ |
| `vllm_cuda13.sqsh` | vllm 0.20.2-nightly (**has** `DeepseekV4ForCausalLM`) | 5.7.0 (no `deepseek_v4`) | ❌ config parse |

The checkpoints declare `"model_type": "deepseek_v4"` / `"architectures": ["DeepseekV4ForCausalLM"]` and ship **no** remote modeling code / `auto_map`, so `--trust-remote-code` is a no-op — config parsing always routes through `transformers.AutoConfig`, which must recognize `deepseek_v4` natively.

## Fix: official-release vLLM image

No source/nightly builds needed — both pieces exist in **stable releases** (confirmed on-cluster):

- **vLLM 0.21.0** — ships the `DeepseekV4ForCausalLM` implementation (release notes: DeepSeek-V4 PP, ROCm, reasoning-effort).
- **transformers 5.9.0** — registers the `deepseek_v4` model_type (`deepseek_v2/v3/**v4**/vl/vl_hybrid`). vLLM 0.21.0 itself pulls transformers 5.7.x, so the Dockerfile pins 5.9.0 on top.

The examples switch from sglang → vLLM accordingly (`--model`, `--tensor-parallel-size`, mirroring the existing `Mistral-Large-3-675B-…-vllm.sh`).

## Status — both examples verified serving on Clariden (2026-05-29) ✅

CI built `images/vllm_deepseek_v4` → `…/ci/vllm_deepseek_v4.sqsh`; both examples were relaunched and serve. `DeepSeek-V4-Pro` returns coherent completions (`"The capital of Switzerland is Bern…"`, `system_fingerprint: vllm-0.21.0-tp8-pp2`).

Three further serving constraints surfaced during validation (all now baked into the examples / env toml, and documented inline in each script header):

1. **`--kv-cache-dtype fp8`** — V4's only attention backend (FlashMLA-Sparse) stores the MLA latent KV exclusively in `fp8_ds_mla`; the default `auto` asserts.
2. **TP must stay 128-aligned** — fp8 weights are 128×128 block-quantised, so `--tensor-parallel-size 16/32` breaks (`input_size_per_partition=192 not divisible by block_k=128`). Use **TP=8** and reach all GPUs via pipeline parallel: V4-Pro `TP8×PP2` (16 GPUs), Base `TP8×PP4` (32 GPUs).
3. **Warmup / JIT timeout** — requests JIT-compile the sparse-attention kernels the first time each *new shape* is seen (not just the first request — it took down both replicas on later prompts). The multi-minute compile overruns vLLM's execute-model RPC deadline (`RPC call to sample_tokens timed out`); raised `VLLM_EXECUTE_MODEL_TIMEOUT_SECONDS=900` in the env toml (pre-warming shapes at startup is a sturdier alternative).

Static checks (shellcheck / hadolint / taplo) pass.


robmsmt:
We build the container image successfully, it loads the model successfully. It starts and takes a request, however when we run it it gets part way through a response and crashes. PySpy points to NCCL deadlock due to TP ... not worth spending lots of time in it since newer versions of vllm that contain new transformers will likely fix this for free

PR #142, opened 2026-05-22
Add Replica-specific Health Check

### LINKED ISSUE #141: Replica-specific Healthchecks
We to enhance the health-check of the models, especially, for the multi-replica models. For now, the health check is done via an end-to-end test as below.

```python
_HEALTH_CHECK_URL = "https://api.swissai.svc.cscs.ch/v1/chat/completions"
_MESSAGE = {"role": "user", "content": "Say hello."}
_TIMEOUT_SECONDS = 10


async def check_model_health(model_name: str, api_key: str) -> ModelHealth:
    try:
        async with httpx.AsyncClient() as client:
            response = await client.post(
                _HEALTH_CHECK_URL,
                headers={
                    "Content-Type": "application/json",
                    "Authorization": f"Bearer {api_key}",
                },
                json={
                    "model": model_name,
                    "messages": [_MESSAGE],
                    "stream": False,
                },
                timeout=_TIMEOUT_SECONDS,
            )
        if response.is_success:
            return ModelHealth.HEALTHY
        return ModelHealth.NOT_RESPONDING
    except (httpx.TransportError, httpx.TimeoutException):
        return ModelHealth.ERROR
```

The problem now is that when we launch a multi-replica model, even if one of them become healthy and responsive, then the health-check will pass. What we should do, is to separately run the healthchecks for the replicas. In the discussion I have with @xzyaoi, we found out that we can use OpenTela directly to do so. We can fetch the list of models from bootstrap (`http://148.187.108.178:8092/v1/dnt/table`). We then can find our models by filtering them out by `served_model_name`. Using the keys (some hash-like things, e.g., `/QmeYmbDUubAg2KCxo3sBLxS2Eity1H5Hsxx4CXM5eYD4cP`), And then hit the specific replica with `http://148.187.108.178:8092/v1/p2p/QmeYmbDUubAg2KCxo3sBLxS2Eity1H5Hsxx4CXM5eYD4cP/v1/_service/llm/v1/chat/completions` like below:

```bash
curl http://148.187.108.178:8092/v1/p2p/QmeYmbDUubAg2KCxo3sBLxS2Eity1H5Hsxx4CXM5eYD4cP/v1/_service/llm/v1/chat/completions \
  -H "Authorization: Bearer apikey" \
  -H "Content-Type: application/json" \
  -d '{
    "model": "Qwen/Qwen3.5-397B-A17B-aWBl",
    "messages": [
      { "role": "system", "content": "You are a helpful assistant." },
      { "role": "user", "content": "What is the capital of France?" }
    ],
    "temperature": 0.7
  }'
```

Where `apikey` is the CSCS API Key which is saved in the InitConfig of SML.

The twist here is that the IP address is not hittable from outside of the internal network. So, when someone is using FirecREST to launch a model, they cannot perform the health check on their local machine. We should submit a job in launcher for the health check and retrieve the report.

It is good to keep both test. Both end-to-end to ensure the model is hittable through `https://api.swissai.svc.cscs.ch` and the replica-specific test to insure all of the replicas are alive.

This should be used for the TUI and the integration tests.

AryanAhadinia:
# Replica-specific health checks + per-replica logs

Screenshot 2026-05-27 at 14 20 16

Closes #141.

## Motivation

The end-to-end health check hits the public gateway (`api.swissai.svc.cscs.ch`), and OpenTela load-balances randomly across peers — so for a multi-replica launch it only ever proves that **one** replica is alive. A replica that is hung or crashed can go unnoticed as long as some other replica answers. This adds a check that confirms **every** replica is healthy, surfaced in the TUI and asserted in the integration tests, while keeping the existing e2e check.

## How it works

The model's **own SLURM job** runs the health checker — no separate helper jobs, no OpenTela bootstrap, and **no API key** in any submitted script.

- A stdlib-only checker (`assets/replica_health_checker.py`) is embedded into the rendered `master.sh` and runs on the batch node, which shares the job's internal network.
- Every 30s it probes each replica's framework `/health` **directly by node IP** (the master already discovers each replica's head IP) and records the last time each replica was seen healthy.
- It writes a JSON report **atomically** (`tmp` file + `os.replace`) to `~/.sml/logs/<job_id>/replica_health.json`, so a concurrent reader — a local read or a FirecREST download — never sees a half-written file (no lock needed; safer than `flock`, which a remote FirecREST read can't participate in anyway).
- The CLI reads that report via a new `Launcher.read_job_file` primitive (local read for SLURM, FirecREST download otherwise). A replica that hasn't been healthy yet reads as `NOT_DEPLOYED` (still starting) rather than a red `ERROR`; a replica that was healthy and then fails surfaces the real `NOT_RESPONDING`/`ERROR`.

## TUI

- New **Replica Health** panel: per-replica node rank, node IP, health, and last heartbeat, with a `healthy/total` summary. Re-rendered every 0.5s so the "Ns ago" heartbeat ticks smoothly between the 30s data refreshes.
- Health monitoring is decoupled from the e2e gateway probe (separate loops), so a slow gateway probe never delays replica/job/log updates.

## Per-replica logs

- Each replica/router `srun` now writes its own `--output`/`--error` (`replica_<r>.out/.err`, `router.out/.err`); `log.out`/`log.err` carry only the master's orchestration output (stdout/stderr are also no longer merged).
- The log viewer is now **tab-per-source** (Master · Replica 0 … Replica N-1 · Router), each with stdout/stderr sub-tabs. Only the **active tab's** logs are fetched each cycle, so log traffic stays bounded regardless of replica count.

## Testing

- Unit: report parsing, in-job checker (health classification, peer-id resolution, last-healthy freezing, startup `NOT_DEPLOYED`, atomic write), log-source mapping, replica panel rendering, and a headless `run_test` of the per-source tabs. Rendered `master.sh` passes `bash -n` + `shellcheck`.
- Integration (`test_firecrest_launcher`): for `replicas > 1`, additionally asserts every replica reports healthy with the full expected count.
- `ruff`, `ruff format`, `mypy --strict` all clean.

## Notes / follow-ups

- **Peer id** is no longer shown in the TUI. The in-job checker still resolves it best-effort (via `srun … /v1/self`, since OpenTela's DNT table carries no internal IP for cluster nodes) and keeps it in the report; that resolution can be dropped if peer id is never needed.
- For multi-node replicas (`nodes_per_replica > 1`), the TUI shows one tab per replica (its head's log, where the framework/HTTP runs); follower nodes write `replica_<r>_node<k>.*` files for manual inspection but don't get their own tabs.
- MCP server is unchanged (scoped to TUI + integration tests).


AryanAhadinia:
The problem with the current implementation is that it submits a job for every health-check. This is very costly and resource consuming. We should fix this.

Suggestion: to run a daemon in the main job to periodically run health check and drop the logs in a file, while the CLI fetch them.

AryanAhadinia:
The importance of this PR is especially escalated due to the bug found in #143.

xzyaoi:
I felt you might want two checks for this purpose:
1. Node:port check: this is for jobs without OpenTela. Since this is a path in SML I assume it needs to be supported as well
2. Check through the OpenTela API (use the URL with peer ID in it to directly call that engine) to check if it is online in the OpenTela network. 

Once the above two checks passed, the node should appear in the serving api’s /models endpoint.

robmsmt (approved):
Looks fine, also the security issues SONAR raises are non-issues for internal node communication imo. Please mark them as not important

PR #145, opened 2026-05-27
Hardcode telemetry endpoint and improve init UX

robmsmt:
## Summary
- **Telemetry endpoint is no longer user-configurable.** The `telemetry_endpoint` wizard step and `--telemetry-endpoint` flag are gone; URL is now a `TELEMETRY_ENDPOINT` constant in `launch_args.py`. Stored `telemetry_endpoint` values in existing `~/.sml/config.yml` are ignored on load (no migration needed).
- **Init wizard polish.** New `intro` field on `_ResolvableConfiguration` renders multi-line guidance before the prompt; URLs in the intro are auto-clickable in modern terminals (iTerm2, kitty, recent Terminal.app).
  - Launcher choice descriptions now point at the FirecREST setup guide and clarify what SLURM mode assumes.
  - FirecREST credential prompt links to the CSCS Developer Portal.
  - CSCS Serving API Key prompt links to `serving.swissai.svc.cscs.ch`.
- **Rename "CSCS API key" → "CSCS Serving API Key"** in all user-visible text (prompts, docs). Storage field `cscs_api_key` and env var `SML_CSCS_API_KEY` are unchanged, so existing configs and keyring entries keep working — no re-init required.
- **Post-launch output adds a `Tail:` line.** Per-launcher hint:
  - SLURM: `tail -f ~/.sml/logs/<id>/log.out`
  - FirecREST: `ssh <host> tail -f ~/.sml/logs/<id>/log.out` (with a note to substitute the user's SSH alias, since we can't reliably guess it)
- **`Launcher.get_log_dir` now returns `~/.sml/logs/<id>`** for both launchers. Previously the absolute `/users/<name>/...` path could be misread as the laptop user's home when launching via FirecREST.

## Test plan
- [x] `uv run pytest tests/unit` — 193 passing
- [ ] Manual: `sml init` (firecrest) — confirm new intros render with clickable links in your terminal
- [ ] Manual: `sml` end-to-end launch on SLURM — confirm `Tail:` shows the direct `tail -f` line
- [ ] Manual: `sml advanced` end-to-end launch via FirecREST — confirm `Tail:` shows the `ssh <host>` placeholder

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -59,7 +59,7 @@ Set `SML_DEBUG=1` to include local variables in crash tracebacks:
 export SML_DEBUG=1
 ```
 
-> **Warning:** `SML_DEBUG=1` may expose secrets (CSCS API key, FirecREST credentials) in crash output. Don't share terminal output captured with this flag.
+> **Warning:** `SML_DEBUG=1` may expose secrets (CSCS Serving API Key, FirecREST credentials) in crash output. Don't share terminal output captured with this flag.
 
 By default, locals are stripped from crash reports.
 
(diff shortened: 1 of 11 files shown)


robmsmt:
@AryanAhadinia most of the changes here are cosmetic, therefore I think we should ignore the test cov

AryanAhadinia (approved):

PR #150, opened 2026-05-29
Configure max tokens for model health check

burakemirsezen:
I spun up a model that was initialized with random weights. Since it's random it doesn't output the eos token when asked Say Hello and thus keeps generating a response until it timeouts. Adding a max token to the request fixes it by cutting the answer short even though the eos is not there.

### DIFF
--- a/src/swiss_ai_model_launch/cli/healthcheck/checker.py
+++ b/src/swiss_ai_model_launch/cli/healthcheck/checker.py
@@ -3,7 +3,7 @@ import httpx
 from swiss_ai_model_launch.cli.healthcheck.model_health import ModelHealth
 
 _HEALTH_CHECK_URL = "https://api.swissai.svc.cscs.ch/v1/chat/completions"
-_MESSAGE = {"role": "user", "content": "Say hello."}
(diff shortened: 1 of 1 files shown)


src/swiss_ai_model_launch/cli/healthcheck/checker.py
```diff
@@ -5,6 +5,7 @@
 _HEALTH_CHECK_URL = "https://api.swissai.svc.cscs.ch/v1/chat/completions"
 _MESSAGE = {"role": "user", "content": "Say hello."}
 _TIMEOUT_SECONDS = 10
+_MAX_TOKENS = 16
```
AryanAhadinia, line 8: Good idea to also keep the answer shorts to not consume resource. I would like to suggest to change the prompt to instruct the model to keep the answer short itself. Something like: `Say the word "Hello". Nothing more."`.

AryanAhadinia (dismissed):
Conditional to resolution of the opened threads either by addressing them or a justification to ignore them.

AryanAhadinia (approved):

PR #152, opened 2026-06-01
Enable multi-platform build support for vllm_apertus_1.5 Docker image

subnet-dev:
Currently, the `vllm_apertus_1.5` Dockerfile is effectively single-platform:
- `TORCH_CUDA_ARCH_LIST` is hard-coded to `"9.0a"`, limiting GPU support.
- The pre-compiled VLLM wheel URL is hard-coded to the **aarch64** flavour, so the image cannot be built on **amd64/x86_64** hosts.
- Python path references are hard-coded to `3.12`.

This MR makes the image truly multi-platform.

## What this MR does

1. **Adds `TARGETARCH` build-arg** so Docker BuildKit can inject the target platform (`amd64` / `arm64`).
2. **Expands `TORCH_CUDA_ARCH_LIST`** to cover a broad range of NVIDIA compute capabilities (`7.5` … `12.0+PTX`).
3. **Selects the correct VLLM wheel at build time** based on `TARGETARCH` (`x86_64` vs `aarch64`).
4. **Adds `--mount=type=cache`** for `apt`, `pip` and `uv` layers to speed up incremental rebuilds.
5. **Switches `uv pip install` torch** from `--index-url` to `--extra-index-url` to avoid masking other package indexes.
6. **Replaces hard-coded `python3.12` paths** with `${PYTHON_VERSION}` for future flexibility.

## Testing notes

- The image should now build on both `linux/amd64` and `linux/arm64` platforms when using `docker buildx build --platform …`.
- Cache mounts require BuildKit (`DOCKER_BUILDKIT=1` or `docker buildx`).

## Breaking changes

None all new values have sensible defaults and the `TORCH_CUDA_ARCH_LIST` default preserves broad compatibility.

### DIFF
--- a/images/vllm_apertus_1.5/Dockerfile
+++ b/images/vllm_apertus_1.5/Dockerfile
@@ -5,8 +5,10 @@ ARG MAX_JOBS=8
 ARG CUDA_VERSION_PATH=cu130
 ARG PYTHON_VERSION=3.12
 ARG VLLM_MAIN_CUDA_VERSION=13.0
+ARG TORCH_CUDA_ARCH_LIST="7.5 8.0 8.6 8.9 9.0 10.0 11.0 12.0+PTX"
+ARG TARGETARCH
 
-ENV TORCH_CUDA_ARCH_LIST="9.0a" \
+ENV TORCH_CUDA_ARCH_LIST="${TORCH_CUDA_ARCH_LIST}" \
     MAX_JOBS="${MAX_JOBS}" \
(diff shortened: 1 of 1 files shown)


AryanAhadinia (commented):
I wonder if you have actually check the build in x64. I think would be nice to do so before merging.

AryanAhadinia (approved):
Approval is conditioned to checking the build in x64 arch.

PR #153, opened 2026-06-02
doc: Add long context checkpoint

dtamayo-nlp:
Add a launcher for the following checkpoint:

```
/capstor/store/cscs/swissai/infra01/apertus_1p5/hf_checkpoints/ap1p5-8b-sft-256k-adam-lr6e-5-constant-128n_5800 
```

The flag `--skip-mm-profiling` was added after a conversation with @Anunay-Yadav ([slack](https://swissai-initiative.slack.com/archives/C0B7DM9PNAE/p1780320008661349?thread_ts=1780318676.373169&cid=C0B7DM9PNAE)) to support multimodality.

### DIFF
--- a/examples/clariden/cli/swiss-ai/1p5-experiments/Apertus-1.5-8B-sft-long-context-it5800.sh
+++ b/examples/clariden/cli/swiss-ai/1p5-experiments/Apertus-1.5-8B-sft-long-context-it5800.sh
@@ -3,18 +3,17 @@
 sml advanced \
   --partition normal \
   --slurm-nodes 1 \
-  --slurm-reservation SD-69241-apertus-1-5-0 \
   --slurm-time 12:00:00 \
   --serving-framework vllm \
   --slurm-environment src/swiss_ai_model_launch/assets/envs/vllm_apertus_1.5.toml \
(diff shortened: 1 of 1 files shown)


robmsmt:
thanks @dtamayo-nlp , a couple of questions:

1. `/capstor/scratch/cscs/dtamayomela/tokenizer_fixed` - scratch has limited life time, can it be moved to permanent location?
2.   I think we should remove `--slurm-reservation SD-69241-apertus-1-5-0` for these scripts it makes more sense to assume the user has set SML_RESERVATION as an env var. This way in a months time when the reserveration changes the script will still work and not error.
3. If long context is applied- `--max-model-len 8192` can be increased?

dtamayo-nlp:
Script updated with the requested modifications.

> 1. /capstor/scratch/cscs/dtamayomela/tokenizer_fixed - scratch has limited life time, can it be moved to permanent location?

Both tokenizers do the same, they were just created with different scripts.

> 2. I think we should remove --slurm-reservation SD-69241-apertus-1-5-0 for these scripts it makes more sense to assume the user has set SML_RESERVATION as an env var. This way in a months time when the reserveration changes the script will still work and not error.

Removed

> 3. If long context is applied- --max-model-len 8192 can be increased?

Increased to 262k.

robmsmt (approved):
LGTM!

PR #154, opened 2026-06-03
added apertus1.5 8b sft dpo example

robmsmt:
see thread: https://swissai-initiative.slack.com/archives/C0A9JJ7C5K6/p1780418221820059


davmels (approved):
The DPO model is indeed the final, best performing one I've got, based on the final SFT checkpoints. 
Tokenizer and max-model-len looks good!

PR #158, opened 2026-06-19
fix(vllm_apertus_1.5): align Dockerfile ENTRYPOINT with upstream vLLM image

subnet-dev:
### Motivation

The `vllm_apertus_1.5` image is intended to run the vLLM OpenAI-compatible server, but it previously defined no `ENTRYPOINT` (or `CMD`). Without this, users must know and manually type the `vllm serve` command on every `docker run`, which is error-prone and breaks parity with the official `vllm/vllm-openai` image.

### Change

Add the standard vLLM instruction at the end of the Dockerfile:

```dockerfile
ENTRYPOINT ["vllm", "serve"]
```

This aligns vllm_apertus_1.5 with the vllm-project/vllm (https://github.com/vllm-project/vllm/blob/main/docker/Dockerfile) Dockerfile, where the vllm-openai target uses the same ENTRYPOINT.

### DIFF
--- a/images/vllm_apertus_1.5/Dockerfile
+++ b/images/vllm_apertus_1.5/Dockerfile
@@ -140,3 +140,5 @@ printf '%s\n' \
     '  /workspace/benchmark-audio-tokenizer (VLLM_APERTUS_AUDIO_TOKENIZER_CODEBASE)' \
     > /etc/motd
 EOF
+
+ENTRYPOINT ["vllm", "serve"]


subnet-dev:
I notice that the other images don't have a clean entrepont configured either. It would be nice to have all the images with a correctly configured entrepont. What do you think?

AryanAhadinia (commented):
Thanks a lot for your contribution. The usage of the docker images in this repo is for preparing the environment for SLURM jobs, and then, the SLURM will run the job shell script on top of the specified docker image. As a result, we are afraid that specifying entry point for the docker image results in failure of SLURM to run the shell script. I checked this with Claude and it also mentioned that the failure is possible. It would be great if you could make sure that the SLURM jobs with this docker image are working fine before merging the PR in.

AryanAhadinia (approved):

AryanAhadinia:
Hello @subnet-dev, Thanks a lot for your contribution. As the image you have changed (`images/vllm_apertus_1.5/Dockerfile`) is no longer in use, I will close this PR, but, in general, we should consider to set entry-point for all of the images we have. I'll create a corresponding issue for this.

PR #161, opened 2026-06-19
Fix SGL Router Registration Issue | Opening Node Terminal in TUI

### LINKED ISSUE #50: Delete `legacy`
After full transition to SML, the `legacy` scripts should and must be deleted.

### LINKED ISSUE #85: Attach to a Running Job
We have non-interactive and "Quit and Resume" option. Hence, it would be nice to have an option to attach to ANY running job.

Idea by @robmsmt.

### LINKED ISSUE #123: Handle Firecrest 5xx issues
Firecrest is sometimes unstable, at the moment it causes a stack trace and does not recover the streaming. We should be able to catch this and handle it appropriately?

### LINKED ISSUE #137: Rename OCF to OpenTela
OCF has been renamed to OpenTela.

### LINKED ISSUE #143: Router is Not Registered as a Peer in OpenTela
While developing #141 and #142, I discovered that SGLang router is indeed not involved in the loop. I found that if a model being deployed with multiple replicas either with SGLang router (`--use-router`) or with OpenTela (OT) built-in router, all of the replicas with registered in OT as peers, and indeed, the router is not registered. Consider the below launches which only differs in using SGL/OT routers.

```bash
sml advanced \
  --slurm-reservation <res> \
  --firecrest-system clariden \
  --partition normal \
  --serving-framework sglang \
  --slurm-replica 8 \
  --tui \
  --slurm-environment src/swiss_ai_model_launch/assets/envs/sglang.toml \
  --framework-args "--model-path /capstor/store/cscs/swissai/infra01/hf_models/models/swiss-ai/Apertus-8B-Instruct-2509 \
    --served-model-name swiss-ai/Apertus-8B-Instruct-2509-$(whoami)-ot-router \
    --host 0.0.0.0 \
    --enable-metrics"
```

```bash
sml advanced \
  --slurm-reservation <res> \
  --firecrest-system clariden \
  --partition normal \
  --serving-framework sglang \
  --slurm-replica 8 \
  --use-router \
  --tui \
  --slurm-environment src/swiss_ai_model_launch/assets/envs/sglang.toml \
  --framework-args "--model-path /capstor/store/cscs/swissai/infra01/hf_models/models/swiss-ai/Apertus-8B-Instruct-2509 \
    --served-model-name swiss-ai/Apertus-8B-Instruct-2509-$(whoami)-sgl-router \
    --host 0.0.0.0 \
    --enable-metrics"
```

In both, 8 peers with the specified _served model name_ appears in dnt table. As I investigated, in both cases, the models are launched with OT in a same way, but SGL router in never being launched with OT.

Image

AryanAhadinia:
# Register the SGLang router on OpenTela and reframe routing as OCF/SGL

## Summary

When a job is launched with the SGLang router, the router itself was never registered on OpenTela. Each replica head advertised the servable `llm` service directly, so OpenTela load-balanced straight across the replicas and bypassed the router entirely — leaving it stale and never carrying traffic. This PR makes the router the single OpenTela `llm` front door, and reframes the user-facing routing choice from a boolean flag into an explicit strategy.

## Router registration

The router now wraps its launch with OCF and advertises `--service.name llm` on the router port, so OpenTela routes external traffic to it. When a router fronts the replicas, the replica heads drop to metrics-only registration: they stay visible on the mesh for metrics and topology, but no longer advertise `llm`, so OpenTela cannot route around the router straight to a replica. Without a router, behavior is unchanged — each head advertises `llm` as before. The telemetry payload's `ocf_service_port` now reflects where `llm` is actually served (the router port when fronted, otherwise the framework port).

## Routing strategy: OCF / SGL

The `--use-router` boolean is replaced by `--router`, which accepts `OCF` or `SGL` and defaults to `OCF`. `OCF` lets OpenTela load-balance across the replica peers on the mesh; `SGL` puts an in-job SGLang router in front of the replicas. This is applied consistently across the three entry points: the `sml advanced` CLI, the `sml preconfigured` guided flow, and the `launch_preconfigured_model` MCP tool. The default `OCF` preserves today's no-router behavior.

## Internal representation

The internal `use_router: bool` is replaced by a `router: Literal["OCF", "SGL"]` type throughout the data models, launchers, framework rendering, CLI, and MCP — the Literal now enforces the contract at type-check and runtime instead of an unconstrained boolean. The job-env key `USE_ROUTER` becomes `ROUTER`. The telemetry payload keeps its existing `use_router` boolean key (derived from the new field) to avoid breaking the backend schema.

## Tests and docs

Added coverage for the new flag mapping, the router-as-front-door registration, and metrics-only heads, and updated all existing tests to the new field. Updated the user-facing docs to the OCF/SGL model. The full unit suite, ruff, mypy, and markdownlint all pass.


docs/adrs/0001-replica-is-the-orchestration-unit.md
```diff
@@ -23,7 +23,7 @@ A **replica** is one independent inference engine instance — one framework pro
 
 - A SLURM job is `replicas × nodes_per_replica` nodes.
 - A replica spans a contiguous set of nodes (`nodes_per_replica`) and is wrapped in OCF on its head node.
-- The in-job framework router (sglang-router today, enabled via `--use-router`) load-balances **across replicas within a single job**, not within them. This is distinct from OpenTela, which is the cross-job p2p mesh that routes between independent jobs/peers; the framework router shapes traffic *inside* one job, OpenTela picks *which* job a request lands on.
+- The in-job framework router (sglang-router today, enabled via `--router SGL`; OCF mesh routing is the default) load-balances **across replicas within a single job**, not within them. This is distinct from OpenTela, which is the cross-job p2p mesh that routes between independent jobs/peers; the framework router shapes traffic *inside* one job, OpenTela picks *which* job a request lands on.
```
robmsmt: why use --router SGL 

i assume this means we use router of sglang. However in other parts of model launch we specify --framework sglang ... I think we should be consistent with the naming... agree?

docs/adrs/0001-replica-is-the-orchestration-unit.md
```diff
@@ -44,13 +44,13 @@ A **replica** is one independent inference engine instance — one framework pro
   
   This is real complexity bringing back exactly the kind of multi-source-of-truth port management we just collapsed. The optimisation it unlocks (DP fanout on a single host) is mostly already handled by sglang's internal scheduler with `--dp-size N`. We defer this until there's concrete demand.
 
-- **`use_router=True` with `replicas=1`.** Validates as an error: there's nothing to load-balance across. (sglang's internal DP gives one HTTP endpoint regardless of how many DP workers it runs.)
+- **`--router SGL` with `replicas=1`.** Silently degrades to no router (served directly), since there's nothing to load-balance across; in interactive mode SGL isn't offered as an option for a single replica. (sglang's internal DP gives one HTTP endpoint regardless of how many DP workers it runs.)
```
robmsmt: as https://github.com/swiss-ai/model-launch/pull/161/changes#r3448625749

docs/faq.md
```diff
@@ -18,7 +18,7 @@ See [How to size a model](sizing.md). The short answer is "enough VRAM to fit we
 
 Yes — by default OCF/OpenTela does random assignment among peers. There is a PR in progress which can change this to different assignment modes: <https://github.com/swiss-ai/OpenTela/pull/4>
 
-In SML, you can pass `--use-router` to put a router in front of N replicas. Without it, you get N independent endpoints with no traffic shaping; this works both with and without OpenTela. OpenTela gives you external access via <https://serving.swissai.svc.cscs.ch> and the API. Without OpenTela (using `--disable-ocf`), the model will not appear there and must be accessed directly from the cluster.
+In SML, you can pass `--router SGL` to put an in-job SGLang router in front of N replicas. With the default `--router OCF`, you get N independent endpoints and OpenTela load-balances across them on the mesh; this works both with and without OpenTela. OpenTela gives you external access via <https://serving.swissai.svc.cscs.ch> and the API. Without OpenTela (using `--disable-ocf`), the model will not appear there and must be accessed directly from the cluster.
```
robmsmt: SGL

docs/initialization.md
```diff
@@ -43,18 +43,35 @@ The FirecREST fields are only required when `--launcher firecrest`. `SML_CSCS_AP
...
+`~/.sml/config.yml` after a successful `firecrest` init looks roughly like:
 
 ```yaml
-launcher: firecrest
-firecrest_url: https://api.cscs.ch/...
-firecrest_token_uri: https://auth.cscs.ch/...
-firecrest_client_id: <secret>
-firecrest_client_secret: <secret>
-cscs_api_key: <secret>
+name: init_config
```
robmsmt, line 51: what is the advantage of nesting this config? What happens if someone has an old config, will it still work?

docs/usage-advanced.md
```diff
@@ -10,31 +10,31 @@ For the guided flow with a curated catalog, use [`sml`](usage-sml.md).
...
+| `--time`                    |                        | Total uptime `HH:MM:SS` (default: `02:00:00`)                                                                                  |
+| `--consecutive`             |                        | Serve a `--time` longer than the per-job cap with a chain of jobs                                                              |
+| `--handover-time`           |                        | Overlap before the previous job ends (default: `03:00:00`)                                                                     |
+| `--max-job-time`            |                        | Per-job cap for chains `HH:MM:SS` (default: `12:00:00`)                                                                        |
+| `--served-model-name`       |                        | Required: pass it here, or include `--served-model-name <name>` inside `--framework-args`. Omitting both aborts with an error. |
+| `--router`                  |                        | Routing: `OCF` (default) or `SGL` (in-job router, replicas > 1)                                                                |
```
robmsmt: OCF should be OpenTela or OTELA

src/swiss_ai_model_launch/launchers/launch_args.py
```diff
@@ -1,11 +1,19 @@
...
 
 from pydantic import BaseModel, Field, model_validator
 
 from swiss_ai_model_launch.launchers.topology import Topology
 
+# Routing strategy across replicas. OCF (default): OpenTela load-balances across
+# the replica peers on the mesh. SGL: an in-job SGLang router fronts the replicas
+# and becomes the served endpoint.
+RouterMode = Literal["OCF", "SGL"]
+ROUTER_OCF: RouterMode = "OCF"
```
robmsmt: as above consider rename to `opentela` / `sglang`

robmsmt (changes requested):
Good change, minor comment on the naming

docs/adrs/0001-replica-is-the-orchestration-unit.md
AryanAhadinia: In case we have one than more replicas, we are using router anyway. Previously, we had `--use-router` flag, for enabling SGL router. In absence of `--use-router` flag, we were falling back to OpenTela router. This naming was a bit problematic because we were using router, even in absence of `--use-router` flag.

So, I renamed it to `--router` with a default of OCF. If user wants to use SGL router instead, they may specify it as `--router SGL`.

Please let me know if you agree with this new convention.

docs/adrs/0001-replica-is-the-orchestration-unit.md
AryanAhadinia: I resolve this thread as the discussions are duplicate to the reference thread. I'll remember to change here if we conclude to rollback in the reference thread.

docs/initialization.md
AryanAhadinia: I didn't changed the implementation at all. The previous doc was not matching to what we had in the code, so, I fixed the doc.

docs/usage-advanced.md
AryanAhadinia: We have a separate issue #137 for renaming OCF to OpenTela, this is the reason I kept using OCF to keep the code consistent and also make the work for Claude easier to implement #137. I can do #137 also in this PR if you want.

src/swiss_ai_model_launch/launchers/launch_args.py
AryanAhadinia: ref: https://github.com/swiss-ai/model-launch/pull/161/changes#r3448639059
Will close the thread.

robmsmt (dismissed):
LGTM!

PR #166, opened 2026-06-28
Add Long Context Apertus 70b Test Checkpoint

AryanAhadinia:


AryanAhadinia:
@dtamayo-nlp, could you please confirm the maximum lengths of the models? Also, would be great if you could take a look to the framework args. I have opted the one we have used for Apertus 1.5 8B.

dtamayo-nlp (changes requested):
`--max-model-len 100000` should be `--max-model-len 262144`. The rest looks good to me.

dtamayo-nlp (approved):
After checking the models online, everything else seems correct.

PR #173, opened 2026-07-10
Robmsmt/vllm v0.19 bos token

robmsmt:
# Fix Apertus double-BOS: chat template owns the BOS token

## The bug

Apertus 1.5's chat template renders `{{ bos_token }}` itself, but vLLM 0.19's default tokenization also added special tokens, so some paths fed the model `<s><s>...` — a double BOS it was never trained on. Instrumented probing (a temporary patch-layer image with print statements + prompt-token-id inspection) showed exactly which paths were broken:

| Path | Before | Why |
|---|---|---|
| Offline `LLM.chat()` (evals, RL rollouts) | ❌ `[1, 1, ...]` | `renderers/base.py` collapses chat/completion defaults into `mm_processor.info.default_tok_params` for multimodal models, which hardcoded `add_special_tokens=True` |
| Server `/v1/chat/completions`, **text-only** | ✅ single BOS | OpenAI layer passes explicit `add_special_tokens=False` per request |
| Server chat **with image/audio** | ❌ `[1, 1, ...]` — **failing in production now** | `ApertusMultiModalProcessor.apply()` decodes the already-tokenized prompt to text and re-encodes it with tokenizer defaults, re-adding BOS |
| `llm.generate()` / `/v1/completions` raw prompts | ✅ single BOS | tokenizer owns BOS on raw paths |

## Changes in `swiss-ai/vllm` — branch `apertus-bos-fix` (1 commit, `b5cb2818c`)

Based directly on `09eae2122` (the exact commit the published image was built from). One file, `vllm/model_executor/models/apertus.py`, +28 lines, two changes:

1. **`ApertusProcessingInfo.get_default_tok_params()` override** — when the tokenizer carries a chat template, return `add_special_tokens=False`: the template owns the BOS. Checkpoints without a chat template keep the default `True` so raw prompts still get their BOS. Same pattern as upstream vllm-project/vllm#39842 (Gemma 4) and the existing ovis/ultravox/paligemma overrides. → fixes offline `LLM.chat()`.
2. **`ApertusMultiModalProcessor.apply()`** — when the prompt arrived as token ids (already tokenized by the renderer per the request's flags), set `add_special_tokens=False` for the re-encode unless the request explicitly said otherwise. Apertus-specific: only its processor does the decode→re-encode round-trip. → fixes server image/audio chat.

## Changes in `swiss-ai/model-launch` — branch `robmsmt/vllm-0-19-bos` (PR diff: 2 files)

1. **`images/vllm_apertus_1.5/Dockerfile`** — pin bumped from `09eae2122` to `apertus-bos-fix` @ `b5cb2818c` (identical code plus the fix; no other dependency changes).
2. **`.github/scripts/build_image.py`** — CI builds now use `podman build --format docker` so `SHELL` instructions are honored (OCI format silently ignores them; ported from `AA/inf_vllm_buildfix`).

(History also contains the since-removed `images/vllm_apertus_1.5_tokdebug/` diagnostic image — net-zero in the PR diff, so CI won't build it.)

## Verification

- From-scratch local build of the pinned Dockerfile (`vllm==0.19.1.dev11+gb5cb2818c`) via eth-cscs/local-registry on clariden, served with `Apertus-1.5-8B-SFT-RL-DPO-SDPO` (checkpoint validated against canonical tokenizer checksums via apertus-omni-tokenizer `validate_model.sh`).
- **quick-model-tests: 58/69.** All model-fixable `special_tokens` BOS checks green, including `bos_single_in_mm_chat{,_audio}` which fail against production today. The two remaining BOS reds (`/completions` with pre-rendered prompts) are vLLM request-schema defaults — clients must send `"add_special_tokens": false` — not model behavior. Other non-passes are pre-existing model/template traits (no parallel tool calls, 8B tool-name leak, 400 on multiple system messages).
- Offline probes: `llm.chat()` went from `[1, 1, 61...]` to `[1, 61...]`; server text paths byte-identical before/after (no regression).

## Rollout notes

- Merging the PR builds `ghcr.io/swiss-ai/vllm_apertus_1.5` (arm64+amd64) and **moves `:latest`** to the fixed build; digest-pinned deployments (e.g. the bosfix layer's `@sha256:44bbfc...`) are untouched until re-rolled.
- Tell the eval team: offline `llm.chat()` runs were silently double-BOSed until now, so post-fix eval numbers may shift (likely upward).

Fix in vllm repo: https://github.com/swiss-ai/vllm/commit/b5cb2818c8b8968e1278d071511b85bc366cf834


RaphaelKreft:
@robmsmt can we have the PR for vllm here linked as well for future reference?

AryanAhadinia (approved):

PR #174, opened 2026-07-10
Publish PR image builds to a pr-<N> channel, not :latest

AryanAhadinia:
## Problem

Docker images built in a PR were published to `ghcr.io/swiss-ai/<image>:latest` and to the shared capstor path `container-images/ci/<image>-<arch>.sqsh`, overwriting whatever main had published.

The sqsh path is the more serious half: the env TOMLs point SLURM launches straight at it, so an open PR could swap the image out from under a launch running off main.

## 1. PR builds get their own channel

A **release channel** is threaded from `ci.yml` through `build_image.py`: `latest` for main, `pr-<number>` for pull requests.

| Artifact | Before (any build) | After (PR #42) |
|---|---|---|
| GHCR per-arch | `:latest-arm64` | `:pr-42-arm64` |
| GHCR manifest | `:latest` | `:pr-42` |
| capstor sqsh | `ci/<image>-arm64.sqsh` | `ci/pr-42/<image>-arm64.sqsh` |

PR builds still publish, so they can be pulled and tested — they just no longer overwrite `latest`. Merging to main rebuilds and republishes under `:latest`.

Two scratch paths that PR and main builds shared are now channel-scoped too: the SLURM build context under `~/.sml/image-builds/` and the intermediate `$SCRATCH/*.sqsh`.

### The build cache was the trap

The key was `build-<image>-<arch>-<hashFiles>`, with no channel in it. Changing only the tags would mean a PR that built an image and then merged unchanged hits that cache on main, skips the build, and **never pushes `:latest`** — the same bug, one layer down. The key now leads with the channel.

## 2. `:latest` now means "built from main"

`workflow_dispatch` resolved to the `latest` channel from **any** branch, so a manual dispatch from a feature branch could publish unmerged code as `:latest` — and since its `image` input defaults to empty, it could republish *every* image at once. The channel step now gates on `github.ref` and refuses to publish from anything but `refs/heads/main`.

This removes the ability to dispatch a test build from a feature branch. PR builds cover that case via `pr-<N>`.

## 3. Env TOMLs resolve `{arch}` at launch

Since #157 CI writes `<image>-arm64.sqsh` / `<image>-amd64.sqsh`, but four of the five env TOMLs read the un-suffixed `<image>.sqsh`.

**Nothing was broken yet**, and it's worth being precise about why: `detect-changes` builds only *changed* image dirs, and `k6`, `sglang_cuda13`, `vllm_cuda13`, `sglang_0.5.10_rocm` haven't changed since the arch split. The un-suffixed files on capstor (May 4, Jun 8) are their last real builds — exactly what the TOMLs read. The trap springs on the *next* edit to any of those dirs: CI writes arch-suffixed artifacts, the TOML keeps reading the stale un-suffixed file, and the change silently never reaches a launch.

A single env TOML has to serve both clusters, and the launcher **never knows the target arch** — it is resolved on the node from `uname -m` (`_render_arch_detection`). So the batch script now substitutes an `{arch}` placeholder into a per-job copy of the env file and hands `srun` the resolved path:

```
image = ".../vllm_apertus_1.5-{arch}.sqsh"   →   -arm64.sqsh on Grace, -amd64.sqsh on x86
```

TOMLs with no placeholder pass through untouched, so pinned paths keep working. A missing image now fails with the path that was actually wanted, instead of an opaque pyxis error deep inside `srun`.

**Only `vllm_apertus_1.5.toml` is switched over here** — it is the one image with both arch artifacts on capstor today. See the rollout note below.

## Cleanup

`cleanup-pr-images.yml` fires on `pull_request: closed` and drops the three `pr-<N>` GHCR tags, the capstor directory, and the build contexts.

Deleting an *org*-owned package version may need admin on the package, which `GITHUB_TOKEN` may not have. The step reads an optional `GHCR_DELETE_TOKEN` secret (a PAT with `delete:packages`) and falls back to `GITHUB_TOKEN`, with `continue-on-error` so leftover tags never fail a merge. **Watch the first close of a PR that actually built an image — if the delete 403s, add the PAT.**

## Rollout, in order

The remaining TOMLs cannot be switched to `{arch}` until their arch-suffixed artifacts exist, and merging this PR does not create them (it changes no `images/` dir, so nothing rebuilds):

1. Merge this PR.
2. Run `workflow_dispatch` on **main** with an empty `image` input to rebuild all five images, producing `<image>-arm64.sqsh` / `<image>-amd64.sqsh`.
3. Follow-up PR: switch `vllm.toml`, `sglang.toml`, `sglang_rocm.toml`, and the k6 loadtest default to `{arch}`.

Doing step 3 before step 2 would point every launch at a file that does not exist.

## Testing

`ruff check` / `ruff format` / `mypy src` clean; 378 unit tests pass; the generated shell block is `shellcheck` clean.

Existing tests don't execute the generated shell, so I ran the rendered block under bash directly:

- templated + `SML_ARCH=arm64` → resolves to `-arm64.sqsh`;
- templated + `SML_ARCH=amd64` → resolves to `-amd64.sqsh`;
- no placeholder → file passed through untouched, exit 0;
- placeholder resolving to a nonexistent image → exits 1 with the wanted path.

Also verified the batch script runs under `set -euo pipefail`, so a failed `sed` aborts the job rather than launching a wrong image, and that `environment` is an absolute remote path by render time, so the resolved copy lands in the writable working dir.

Channel guard rejects `../../etc`, `pr-42; rm -rf /`, `main`, and empty. The deletion `jq` filter matches `pr-4` / `pr-4-arm64` and leaves `pr-42` alone — the first draft used `startswith`, which would have deleted PR #42's image when PR #4 closed.

Not exercised against real infrastructure: the FireCREST `rm` calls, the GHCR deletion API, and an actual two-cluster launch through the `{arch}` path.

https://claude.ai/code/session_014yyiEE7oEK4HbuWEh6iSfA

### DIFF
--- a/.github/scripts/build_image.py
+++ b/.github/scripts/build_image.py
@@ -2,6 +2,7 @@
 
 import asyncio
 import os
+import re
 import sys
 import tempfile
 import time
@@ -11,6 +12,10 @@ from textwrap import dedent
 import firecrest as f7t
 
 _CAPSTOR_IMAGES = "/capstor/store/cscs/swissai/infra01/container-images/ci"
+_RELEASE_CHANNEL = "latest"
+# Anything outside this set lands in a filesystem path and a registry tag, so
+# it must not contain path separators or shell metacharacters.
+_CHANNEL_RE = re.compile(r"^(latest|pr-\d+)$")
 _POLL_INTERVAL = 60
 _TIMEOUT = 4 * 3600
 _TERMINAL_STATES = {
@@ -23,9 +28,22 @@ _TERMINAL_STATES = {
 }
 
 
+def sqsh_path(image_name: str, arch: str, channel: str) -> str:
+    """Where a build's squashfs lands on the shared capstor store.
+
+    The release channel keeps its historical flat path, because the env TOMLs
+    and other consumers point straight at it. Pre-release channels get their
+    own subdirectory so a PR build can never overwrite what main published.
+    """
+    if channel == _RELEASE_CHANNEL:
+        return f"{_CAPSTOR_IMAGES}/{image_name}-{arch}.sqsh"
+    return f"{_CAPSTOR_IMAGES}/{channel}/{image_name}-{arch}.sqsh"
+
+
 def _build_slurm_script(
     image_name: str,
     arch: str,
+    channel: str,
     account: str,
     partition: str,
     reservation: str | None,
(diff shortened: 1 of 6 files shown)


robmsmt (approved):
LGTM, hard to say with these CI files, sometimes better to just run it.

In general i think we should document what triggers a build and where... maybe in docs or images/readme.md

Issue #34, opened 2026-03-27
Add Model Download


AryanAhadinia:
Downloading and storing the model is now a separated step. It should be an integral part of the model launch process and must be done transparently.

robmsmt:
The problem is that those outside infra01 likely cannot run this. 

Another approach that doesn't have this problem- would be that we have a service that runs, and responds if model exists...  If not can download the model. But then this endpoint needs to be secured / whitelisted etc.

AryanAhadinia:
For people outside of infra01, it would not introduce a new problem since the download would be triggered only if the model is not already available. If the model is already available, then the download would not be triggered.

However, there is a decision we have to make in connection. Toward #122 resolution, we may restrict the mounting as read-only, but if we do so, we cannot integrate the downloading in the same job and we have to submit a separate job for downloading the model. In security PoV, this will give us a better isolation since model download job is non-user-interactive, in contrast, model launch job is highly exposed to the users, so it would be a good idea to not grant write access to model launch submission.

In summary, below are the options:
1. Having separate jobs for downloading the models, and grant read-only access to launch job.
2. Having an integrated jobs with `rw` access.

@robmsmt, what do you think?

PR #195, opened 2026-08-04
feat: Add servekit integration for fast weight loading on sglang with GLM5.1-FP8 and GLM4.7

youssef62:
This PR is a proof of concept for integrating servekit with SML. Its main goal is to support discussions about **when** servekit needs to be used effectively and **how** it should be improved to meet the serving team's requirements.

## Changes

* Integrates servekit's fast loading through a `--servekit-optims` flag that's currently only available for sglang (no vllm) and slurm (no firecrest).
* Add a catalog model (apertus-8b) that uses servekit for demo and benchmark reasons.

## Benchmarks

**GLM5.1-FP8** 

https://github.com/user-attachments/assets/8ec826a0-3555-4d98-8781-bef08b806b26

For a complete evaluation of servekit. See: https://github.com/youssef62/cold-start-elimination-experiments-swissai#3-fast-weight-loading-with-servekit


examples/servekit/prepare.sbatch
AryanAhadinia, line 1: I think, it would be great if we integrate this with download job [#34]. We usually download and cache the models in `capstor` rather than downloading them each time.

The twist here is to find the right TP for the models. We usually estimate the TP based on the number and type (FP32/BF16) of parameters and converge to the final number by trial and error.

Since we are now doing this pre-sharding, the aforementioned trial and error would be costly.

It would be nice if we can find a systematic way to fix parameters such as TP out of a formula. I know this may be a bit out of context for this PR, but as I thought this may affect the usability of this new feature since it will be more expensive for developers to add support for newer models.

Just to give you the bigger picture, we are moving toward publishing SML as a package for community, so it is kind of important for us to keep it simple.

src/swiss_ai_model_launch/assets/models.json
```diff
@@ -84,6 +84,16 @@
     "nodes_per_replica": 2,
     "framework_args": "--tp-size 8 --enable-metrics"
   },
+  {
+    "model": "swiss-ai/Apertus-70B-Instruct-2509-servekit",
+    "framework": "sglang",
+    "environment": null,
+    "nodes_per_replica": 2,
+    "framework_args": "--load-format sharded_state --tp-size 8 --enable-metrics",
+    "pre_launch_cmds": "pip install --quiet git+https://github.com/eth-easl/servekit.git@e9fde0b",
```
AryanAhadinia, line 93: This `pre_launch_cmds` is more for development stage of testing some new model. In general, if we need a dependency globally, we will add it to the docker images we have in `./images`. For now, you can go on with this since it would be hard to build the docker images. Ideally, we can move this to docker images before merging with main.

src/swiss_ai_model_launch/launchers/firecrest_launcher.py
```diff
@@ -112,6 +112,8 @@ def _get_launch_args_from_request(
         self,
         launch_request: LaunchRequest,
     ) -> LaunchArgs:
+        if launch_request.servekit_optims:
+            raise ValueError("servekit_optims is only supported with the direct SLURM launcher, not FirecREST.")
```
AryanAhadinia, line 116: SLURM and FirecREST launcher are both inherited from `Launcher` superclass and they should expose same behavior with different implementation.

FYI, SLURM launcher is used when user is running SML directly on the cluster and FirecREST is for the time that users are running SML on another machine.

FYI, FirecREST is a framework that CSCS has developed to submit SLURM jobs via API to SLURM clusters.

src/swiss_ai_model_launch/launchers/launch_args.py
```diff
@@ -102,11 +102,17 @@ class LaunchArgs(BaseModel):
...
     @model_validator(mode="after")
     def _validate(self) -> "LaunchArgs":
         if not self.disable_metrics and not self.metrics_remote_write_url:
             raise ValueError("Metrics require a remote write URL when metrics are enabled.")
+        if self.servekit_optims and self.framework != "sglang":
+            raise ValueError(
+                "--servekit-optims is only supported with --framework sglang "
+                "(servekit's fast weight loading does not yet support vLLM)."
+            )
         if _PORT_FLAG_RE.search(self.framework_args):
```
AryanAhadinia, line 116: We also want to be symmetric to framework, but we also have other asymmetries regarding frameworks. This is not as serious as the point I mentioned in [this comment](https://github.com/swiss-ai/model-launch/pull/195/changes#r3716649905).

AryanAhadinia (commented):
You are just on the direction. Keep on doing this great job!

AryanAhadinia:
@youssef62, I would like to recommend you to rebase your branch on a regular basis. We are moving forward fast and accumulating the conflicts may get out of control.

examples/servekit/prepare.sbatch
youssef62: > I think it would be great if we integrate this with the download job [https://github.com/swiss-ai/model-launch/issues/34]. We usually download and cache the models in `capstor` rather than downloading them each time.

Adding it to a download job sounds nice. But the issue is that the prepare step depends on the model parallelism config (tp, pp...) which we don't fix at download time, I guess. 

> The twist here is to find the right TP for the models. We usually estimate the TP based on the number and type (FP32/BF16) of parameters and converge to the final number by trial and error.

> Since we are now doing this pre-sharding, the aforementioned trial and error would be costly.

> It would be nice if we could find a systematic way to fix parameters such as TP out of a formula. I know this may be a bit out of context for this PR, but as I thought this may affect the usability of this new feature since it will be more expensive for developers to add support for newer models.

Hmm, I agree that the presharding takes time. Currently, `servekit` is more oriented to speedups during production, not dev. But for development, if we consider that devs can do their (tp, pp, ep) configs experiments as usual and add servekit at the end before production, then it does not slow down development. 

For fixing the (tp,pp,ep) config automatically, it looks a bit orthogonal to what I am working on. Maybe @xzyaoi  has ideas.

src/swiss_ai_model_launch/launchers/firecrest_launcher.py
youssef62: done 642b4ae6a9a02437a8ad9410f9268eab6298489b

src/swiss_ai_model_launch/assets/envs/sglang5.16_base.toml
AryanAhadinia, line 1: As I mentioned before, we maintain our own images and currently I am working on updating our SGL/vLLM images. Before merging this PR, would be great to point to our own SGL image.

youssef62: I added this image because `servekit` needs >=sglang5.14. If this blocks the PR from being merged. Can we have a github issue for your new image and link it to this PR so this one does not get lost? 
What do you think ? @AryanAhadinia @robmsmt

AryanAhadinia (approved):

AryanAhadinia (changes requested):
Could you please rebase to latest main to fix the CI pipeline?

robmsmt (commented):
Went through the diff and through servekit at the pinned commit (0d54a36). Adding detail to Rob's point above that the catalog is the wrong place for this: the entries don't just name a cluster-specific artifact, they change what a launch is allowed to do.

**What servekit writes**

- *Weight artifact on capstor.* Each entry points at `swiss-ai-serving/servekit-artifacts/<model-layout>`, a pre-sharded copy of the checkpoint for one tp/pp layout on one GPU shape (already there: GLM-4.7 662 GB, GLM-5.1-FP8 702 GB, Apertus-8B 16 GB). `servekit launch` **creates the artifact itself if the directory is empty** (`launch.py: _prepare_artifact`; `_writable` walks up to the nearest existing dir and mkdirs). So a launch job can drop a ~700 GB copy under the shared serving tree as a side effect. There is no read-only mode; an unwritable path warns and falls back to the engine's loader.
- *JIT caches beside the artifact.* Since eth-easl/servekit d866545 (in the pin), Triton / TVM-FFI / FlashInfer caches are written into `<artifact>/servekit-caches` by prepare and copied to node-local `/tmp/servekit/caches` on every launch, with `TRITON_CACHE_DIR` etc. pointed there. Also GPU-shape-specific.
- Weights staged into `/dev/shm/servekit` per node, freed after startup.

**Why this doesn't fit model-launch as it stands**

1. **A launch must not write to capstor.** model-launch launches are read-only against the image and the weights; with these entries a launch with a missing or renamed artifact silently rebuilds 700 GB. If servekit is used at all, artifact creation has to be a separate, explicit step and launches must never prepare.
2. **Layout drift is silent.** If tp/pp/ep in `framework_args` stops matching the artifact manifest, servekit warns and falls back to slow loading; the job passes, just slowly. Nothing in the catalog ties the entry to the artifact it depends on.
3. **The container is mutated at job start.** `pre_launch_cmds` pip-installs servekit from a GitHub tarball, and for GLM-5.1 `servekit._patches.sglang_35715` rewrites sglang's `deepseek_v2.py` inside the running container. That needs network at job start and puts code outside the image; it's the opposite of the pinned-image model the rest of the catalog relies on.
4. **A production entry changes under the PoC.** GLM-5.1-FP8 goes from `--tp-size 16 … EAGLE` to tp4/pp4/ep4 + `--load-format sharded_state` + flashinfer + `--context-length 32768` in the same PR, with no separate gate run for that change.

Agree with the comment above: if this stays, it belongs in `examples/` with the prepare step spelled out, not in `models.json` until the catalog knows about clusters and GPU shapes.

src/swiss_ai_model_launch/assets/models.json
```diff
@@ -154,9 +154,24 @@
   {
     "model": "zai-org/GLM-5.1-FP8",
     "framework": "sglang",
-    "environment": null,
+    "environment": "src/swiss_ai_model_launch/assets/envs/sglang5.16_base.toml",
```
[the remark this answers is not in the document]
robmsmt: Serve kit weights are tightly coupled to the architecture/gpu shape... therefore adding it to models.json is the wrong place. It should be in examples only. Unless we make model.json cluster aware.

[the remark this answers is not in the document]
youssef62: > Serve kit weights are tightly coupled to the architecture/gpu shape... therefore adding it to models.json is the wrong place. It should be in examples only. Unless we make model.json cluster aware.

The weights are not coupled to the hardware architecture. In fact, the same weights work on clariden and bristen. They are just coupled to the parallelism setup (dp/tp/pp).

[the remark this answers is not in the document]
robmsmt: agreed, this is what I meant. the size of the VRAM usually dictates the parallelism setup. My point was for Aryan that our models.json should know whether the user has selected an environment and show them the list of models that are supposed to work on it...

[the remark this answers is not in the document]
robmsmt: this is to not block this PR, however i think we should raise an issue to look at it later

PR #198, opened 2026-08-07
Authorization Scheme

AryanAhadinia:
Companion to serving-api's label-based model authorization ([ADR-0001](https://github.com/swiss-ai/serving-api/blob/main/docs/adrs/0001-label-based-model-authorization.md)). A launch now declares who the model is for, as an OpenTela peer label the gateway enforces:

    sml ... --authorization public                        # anyone (default)
    sml ... --authorization private                       # only the launcher
    sml ... --authorization user1@epfl.ch,user2@ethz.ch   # only these users

"private" never reaches the mesh — the gateway cannot know who submitted a job, so the CLI resolves it to the launcher's own email via the Serving API's GET /v1/whoami before submission, and fails the launch rather than falling back to public if that call can't be made. LaunchArgs rejects an unresolved "private" as a guardrail for any path that forgets.

Values are canonicalized (strip, lowercase, dedupe) so two launches meaning the same policy emit the same label, and a served name whose replicas disagree on policy is refused by the gateway for everyone — so a launch that would collide with a running model under a different policy is refused up front instead of taking that model down with it.

Applied on every path that submits a launch: `preconfigured`, `advanced`, both `loadtest` launch paths, and the MCP server's launch tool.

## Launch identity: `launch_id` and `launched_by_email`

A peer label is immutable for the life of the SLURM job, so `--authorization` can only ever set the policy a model *starts* with. serving-api [ADR-0002](https://github.com/swiss-ai/serving-api/blob/main/docs/adrs/0002-post-launch-access-overrides.md) adds a **Manage access** menu that changes a running model's audience without relaunching, and **Reset** to put it back. That needs two facts from the launch that the mesh did not previously carry:

- **`launched_by_email`** — the launcher's platform identity, resolved via `/v1/whoami`. The existing `launched_by` label is `$USER`, a **cluster shell account** (`bdoan`, `dmelikidze`), which cannot be compared against a Serving API key's `owner_email`. Without this label the gateway genuinely cannot tell which platform user launched a model, so it cannot tell whose model it is to manage.
- **`launch_id`** — a UUID identifying *this* launch, generated per `LaunchArgs` so every path that builds one gets a fresh id for free. The gateway records a post-launch access change against this rather than against the served name, so a change can never outlive its job and re-apply to a different launch that later takes the same name.

`launched_by_email` resolution is deliberately **best-effort**, unlike the `private` resolution above. There the email *is* the access policy, so failing to resolve it has to fail the launch. Here it only decides who may edit the policy afterwards, and refusing to launch a public model because the Serving API was briefly unreachable would be the worse trade — so the launch proceeds with a warning, and the model is admin-managed. `whoami` is memoized per API key so a `private` launch still makes one call, not two.

## Other changes

- `sml_version` label, recording which SML rendered a launch — baked in as a literal on the submitting machine, since SML isn't installed on the compute node.
- Framework included in derived served-model names, and examples updated to write served-model names with their namespace.

## Docs

`docs/usage-advanced.md` (Model authorization), `docs/usage-sml.md`, and `docs/glossary.md` now describe both layers: the flag sets the launch-time policy, the web UI changes it afterwards, and a relaunch starts again at whatever the flag says.

## Testing

428 unit tests pass; `ruff` and `markdownlint` clean. New coverage for the launch-identity labels: uniqueness and per-render stability of `launch_id`, that `launched_by_email` defaults to empty rather than to a guess (an invented owner could then change the model's access), and that resolution failure warns without failing the launch.

## Merging plan

Coupled with https://github.com/swiss-ai/serving-api/pull/115, but not merged together — serving-api goes first (Stages 0–1), then this PR is tested against dev and merged (Stages 2–4). Until then the gateway simply sees no `launch_id` / `launched_by_email` labels and reports those models as admin-managed.

### DIFF
--- a/docs/glossary.md
+++ b/docs/glossary.md
@@ -56,6 +56,12 @@ One independent copy of the model (a [DP](sizing.md#parallelism-dp-tp-pp-ep-and-
 
 A SLURM concept — a slot of nodes pre-allocated to a user/group, bypassing the normal queue. Set via `--reservation` (advanced) or `--reservation` (interactive). Optional.
 
+## Model authorization
+
+Who may list and use a served model. Set at launch via `--authorization`: `public` (the default — anyone), `private` (only you), or a comma-separated email list (exactly those users — it does not implicitly include you). It travels to the mesh as an OpenTela peer label and is enforced by the Serving API, which hides unauthorized models from `/v1/models` and answers `403` on inference routes.
+
+`private` never reaches the mesh: the gateway has no idea who launched a job, so SML resolves it to your own email via the Serving API's `/v1/whoami` before submitting. Because OpenTela load-balances a [served-model name](#served-model-name) across every peer advertising it, two launches sharing a name with *different* policies make that name unroutable for everyone — SML refuses such a launch up front. See [Model authorization](usage-advanced.md#model-authorization).
+
 ## Router
 
 A framework-side load balancer (e.g. `sglang-router`) inserted in front of N replicas inside one SLURM job. Enabled via `--router sglang` (the default `--router opentela` skips it and lets OpenTela balance across the replica peers). Orthogonal to [OpenTela](#opentela): the router shapes traffic *within* the job; OpenTela picks *which* job/peer a request lands on.
--- a/docs/glossary.md
+++ b/docs/glossary.md
@@ -56,6 +56,12 @@ One independent copy of the model (a [DP](sizing.md#parallelism-dp-tp-pp-ep-and-
 
 A SLURM concept — a slot of nodes pre-allocated to a user/group, bypassing the normal queue. Set via `--reservation` (advanced) or `--reservation` (interactive). Optional.
 
+## Launch labels
+
+Facts about a launch that SML attaches to its [OpenTela](#opentela) peer, which the gateway reads back — visible per model in `/v1/models_detailed`. Set at peer start and fixed for the job's life:
+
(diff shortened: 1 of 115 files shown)


docs/glossary.md
```diff
@@ -56,6 +56,12 @@ One independent copy of the model (a [DP](sizing.md#parallelism-dp-tp-pp-ep-and-
 
 A SLURM concept — a slot of nodes pre-allocated to a user/group, bypassing the normal queue. Set via `--reservation` (advanced) or `--reservation` (interactive). Optional.
 
+## Model authorization
+
+Who may list and use a served model. Set at launch via `--authorization`: `public` (the default — anyone), `private` (only you), or a comma-separated email list (exactly those users — it does not implicitly include you). It travels to the mesh as an OpenTela peer label and is enforced by the Serving API, which hides unauthorized models from `/v1/models` and answers `403` on inference routes.
+
+`private` never reaches the mesh: the gateway has no idea who launched a job, so SML resolves it to your own email via the Serving API's `/v1/whoami` before submitting. Because OpenTela load-balances a [served-model name](#served-model-name) across every peer advertising it, two launches sharing a name with *different* policies make that name unroutable for everyone — SML refuses such a launch up front. See [Model authorization](usage-advanced.md#model-authorization).
+
```
AryanAhadinia: Write in better wording.

robmsmt:
I want to check we are not missing something from authentik here, I understand how you are implementing this but the auth isn't quite what we specified in the original requirements (won't link it here publicly) 

but agree this is a useful feature

PR #204, opened 2026-08-11
Create New Base Docker Image for vLLM

### LINKED ISSUE #126: vLLM Launch Crashes for Multi-node Models

AryanAhadinia:
## Summary

Adds `vllm_base`: a clean baseline vLLM image built from the upstream release on CUDA 13, with no model-specific patches. The existing `vllm_*` images each carry accumulated fixes for one model or one incident, which makes them hard to reason about and harder to bump. This one is meant to be the maintainable starting point that the others eventually rebase onto.

Along with the image itself, this adds the machinery to keep it honest: a post-build sanity check that runs against the built artifact, build provenance that survives into the `.sqsh`, and a nightly pipeline that proposes dependency bumps as reviewable PRs.

## What's here

**The image** — `images/vllm_base/Dockerfile`

`nvidia/cuda:13.0.3-cudnn-devel-ubuntu24.04`, venv at `/opt/venv`, everything installed with `uv`. The comments carry the reasoning for the parts that are not obvious: why `devel` rather than `runtime` (FlashInfer, tilelang, quack-kernels and humming-kernels all JIT at runtime and need `nvcc`), why `LD_LIBRARY_PATH` puts the host driver libraries first (CUDA Error 803 from the `cuda-compat` stubs), why `flashinfer-cubin` has to be named explicitly (vLLM pins it but excludes it from published wheel metadata, so it is otherwise JIT-compiled on first use in every job), and why the cuDNN/NCCL headers are taken from the wheels rather than the base image.

`PYTHON_VERSION` now actually selects the interpreter — the packages are requested as `python3.12`/`-dev`/`-venv` rather than bare `python3`. Previously the ARG only described whatever the base image's default happened to be, and a base bump would have left `LD_LIBRARY_PATH` silently pointing at a directory that no longer existed.

**Sanity check** — `images/vllm_base/sanity_check.sh`

Runs against the *built* image rather than during the build, between `podman build` and `podman push`, so a failure blocks the release instead of scrolling past in a build log. Driven entirely by `SML_SANITY_*` variables the Dockerfile bakes in, so it is image-agnostic — other `images/` entries adopt it by dropping the file in and declaring a contract.

It asserts the interpreter version, commands on `PATH`, exact installed distribution versions (via `importlib.metadata`, no import needed), that `torch` and `vllm` import, and that the cuDNN/NCCL symlinks resolve. It accumulates failures rather than exiting on the first, so one run shows everything wrong. The GPUs are passed into the check via CDI (`--device nvidia.com/gpu=all`) — importing vLLM initialises platform detection, which needs a live CUDA runtime, so that import doubles as proof the runtime actually comes up.

This replaces the version-printing `RUN` at the end of the Dockerfile, which only printed into a log nobody reads.

**Build provenance** — `SML_IMAGE_BUILD`

`vllm_base pr-204 arm64 <sha> <timestamp>`, injected as build args by CI. Carried in the environment rather than as OCI labels, because `enroot import` produces a bare rootfs and keeps only the image environment (as `/etc/environment`) — labels do not survive into the `.sqsh` the cluster actually runs. For PR builds the recorded SHA is the branch head, not the ephemeral merge commit.

**Nightly pipeline** — `.github/workflows/nightly.yml`, `.github/scripts/nightly_bump.py`

Runs at 02:00 UTC. Resolves the latest versions for images that ship a `nightly.toml`, rewrites the ARGs, builds into a rolling `:nightly` channel, and opens a PR with the diff.

The bump is not a naive "latest everything". vLLM hard-pins the torch trio and `flashinfer-cubin` with `==` in its `requirements/cuda.txt`, so those follow the resolved vLLM release rather than their own latest — bumping them independently makes the resolve unsatisfiable. Pre-releases and fully-yanked releases are filtered out.

A quiet night is a no-op: no branch, no build, no PR. A failing build still opens a PR, as a draft marked `— FAILING` with the sanity-check output inline, because the broken bump is exactly the thing worth seeing; `:nightly` stays pointing at the last good build. Each night closes the previous nightly PR, so there is never a backlog.

`ci.yml` skips its build job for `nightly/*` head refs — the nightly already built and scanned those exact Dockerfiles, and rebuilding under `pr-<N>` would spend a second multi-hour job per arch on identical code. Static checks and integration tests still run on the PR, and the real build happens on merge to `main`.

**Supporting changes**

- `src/swiss_ai_model_launch/assets/envs/vllm_base.toml` — launch env: NCCL/libfabric settings for the Slingshot fabric, and JIT caches pointed at capstor rather than NFS `$HOME` or node-local tmpfs.
- `.github/image-scan-allowlist.txt` — three suppressions for known-benign scan findings, each with the reasoning recorded. The Lob one is worth reading: TruffleHog's verifier treats a 422 from an empty POST body as proof of a live key, so any image shipping third-party Python test suites trips it in the hundreds.
- `_CHANNEL_RE` in `build_image.py` and `scan_image.py` widened to accept `nightly`.

## Verification

`ruff`, `mypy`, `shellcheck`, `hadolint`, `taplo` and `prettier` all pass.

`nightly_bump.py` was tested end to end against a stubbed network: the opt-in marker (an image without `nightly.toml` is left alone), torch correctly following the newly resolved vLLM rather than the current one, pre-release and yanked filtering, the `GITHUB_OUTPUT` contract, and a second run correctly reporting no changes.

`sanity_check.sh` was exercised locally against passing, failing (all six failure modes, including a dangling symlink), and empty-contract inputs.

CDI GPU passthrough was confirmed on a login node: a rootless `podman run --device nvidia.com/gpu=all` sees `/dev/nvidia0..3` inside the container.

## Not yet verified

- **The nightly's live network paths.** `pypi_latest` and `vllm_pins` have only run against stubs — the dev node has no egress. The parsing is tested against real `requirements/cuda.txt` content, but the first live run is the actual proof. Worth a manual `workflow_dispatch` rather than waiting for 02:00.
- **`python3.12` resolving on the CUDA base image's archive.** Confident from how noble is built, but the first CI build is the confirmation.
- **`ray` on `PATH` in the built image.** Listed in `SML_SANITY_COMMANDS`; the other entries are certain. If the sanity check fails there, drop it from the list.
- **GPU visibility on the build partition.** The SLURM job has no `--gres`, so the sanity check relies on the build nodes exposing GPUs. True for Grace; unconfirmed for the amd64 cluster.

## Before merging

- [ ] Add `secrets.NIGHTLY_PR_TOKEN` (a PAT with `repo` scope). Without it the nightly's `open-pr` job fails — a `GITHUB_TOKEN`-authored PR gets no checks at all, which is why a PAT is needed.
- [ ] `vllm_base.toml` points at `.../ci/pr-204/vllm_base-{arch}.sqsh` — this PR's own pre-release channel, which `cleanup-pr-images.yml` deletes when the PR closes. It needs to be the flat `latest` path before merging.
- [ ] Decide whether the nightly should run the secret scan. It is included here to match how `pr-<N>` builds are treated, but it adds up to 90 minutes; the scan also runs on merge to `main` regardless.

Note that vLLM 0.28.0 carries breaking changes: bitsandbytes moved out-of-tree, `calculate_kv_scales` and `override_attention_dtype` removed, and `reasoning_content` dropped from output — that last one is client-visible.

https://claude.ai/code/session_016rs1ZyXZqf1LAGxC19iPxu

### DIFF
--- a/images/vllm_base/Dockerfile
+++ b/images/vllm_base/Dockerfile
@@ -107,9 +107,10 @@ RUN SITE="/opt/venv/lib/python${PYTHON_VERSION}/site-packages" && \
 # between `podman build` and `podman push` (.github/scripts/build_image.py), so
 # a mismatch blocks the release instead of reaching a build log nobody reads.
 #
-# Distributions are checked through metadata; torch and vLLM are also imported
-# -- torch catches a mismatched CUDA ABI, and vLLM initialises platform
-# detection, which proves the CUDA runtime comes up on the build node's GPUs.
+# Distributions are checked through metadata; torch is imported, catching a
+# mismatched CUDA ABI. vLLM is GPU-gated: importing it initialises platform
+# detection and needs a live CUDA runtime, which the arm64 build nodes provide
+# and the amd64 ones do not.
 ENV SML_SANITY_PYTHON="${PYTHON_VERSION}" \
     SML_SANITY_COMMANDS="python uv curl ffmpeg vllm ray" \
     SML_SANITY_PACKAGES="torch==${TORCH_VERSION} \
@@ -119,7 +120,8 @@ vllm==${VLLM_VERSION} \
 flashinfer-cubin==${FLASHINFER_VERSION} \
 transformers==${TRANSFORMERS_VERSION} \
 ray==${RAY_VERSION}" \
-    SML_SANITY_IMPORTS="torch vllm" \
+    SML_SANITY_IMPORTS="torch" \
+    SML_SANITY_GPU_IMPORTS="vllm" \
     SML_SANITY_FILES="${VIRTUAL_ENV}/include/cudnn.h ${VIRTUAL_ENV}/include/nccl.h"
 
 # Provenance in the environment because that is what reaches the cluster:
--- a/images/vllm_base/Dockerfile
+++ b/images/vllm_base/Dockerfile
@@ -1,8 +1,8 @@
 # Baseline vLLM image: upstream release on CUDA 13, no model-specific patches.
 #
-# The base tracks CUDA 13.0.x because the torch and vLLM wheels are cu130
-# builds. The devel variant is required, not incidental: FlashInfer, tilelang,
-# quack-kernels and humming-kernels all JIT-compile at runtime and need nvcc.
+# cu130 matches the torch and vLLM wheels. The devel variant is required, not
+# incidental: FlashInfer, tilelang, quack-kernels and humming-kernels all
+# JIT-compile at runtime and need nvcc.
 FROM docker.io/nvidia/cuda:13.0.3-cudnn-devel-ubuntu24.04
 
 ARG DEBIAN_FRONTEND=noninteractive
@@ -18,22 +18,28 @@ ARG TRANSFORMERS_VERSION=5.15.0
 ARG RAY_VERSION=2.57.0
 ARG FLASHINFER_VERSION=0.6.16.post3
 
+# Build provenance from CI (.github/scripts/build_image.py); "unknown" in a published image means it was not built by the pipeline.
+ARG BUILD_IMAGE=unknown
+ARG BUILD_ARCH=unknown
+ARG BUILD_CHANNEL=unknown
+ARG BUILD_REVISION=unknown
+ARG BUILD_DATE=unknown
+
 ENV PIP_DISABLE_PIP_VERSION_CHECK=1 \
     PIP_NO_CACHE_DIR=1 \
     UV_HTTP_TIMEOUT=500 \
     UV_INDEX_STRATEGY=unsafe-best-match \
     UV_LINK_MODE=copy
 
-# Keep host NVIDIA driver libraries first, so the cuda-compat stubs in
-# /usr/local/cuda/compat are not loaded ahead of them (CUDA Error 803).
+# Keep host NVIDIA driver libraries first, so the cuda-compat stubs in /usr/local/cuda/compat are not loaded ahead of them (CUDA Error 803).
 ENV LD_LIBRARY_PATH=/usr/local/nvidia/lib:/usr/local/nvidia/lib64:/usr/lib
 
 SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 
 WORKDIR /opt
 
-# libibverbs/libnuma back the RDMA and NUMA paths on Grace nodes; ffmpeg and
-# libsndfile1 decode audio/video inputs; curl serves the router health checks.
+# DL3008 forces pinning versions explicitly in apt command. 
+# We ignore this for apt installed packages because we want the latest images for security reasons.
 # hadolint ignore=DL3008
 RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
     --mount=type=cache,target=/var/lib/apt,sharing=locked \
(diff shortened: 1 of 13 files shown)


images/vllm_base/Dockerfile
```diff
@@ -0,0 +1,98 @@
...
+# /usr/local/cuda/compat are not loaded ahead of them (CUDA Error 803).
+ENV LD_LIBRARY_PATH=/usr/local/nvidia/lib:/usr/local/nvidia/lib64:/usr/lib
+
+SHELL ["/bin/bash", "-o", "pipefail", "-c"]
+
+WORKDIR /opt
+
+# libibverbs/libnuma back the RDMA and NUMA paths on Grace nodes; ffmpeg and
+# libsndfile1 decode audio/video inputs; curl serves the router health checks.
+# hadolint ignore=DL3008
```
AryanAhadinia, line 46: Should be fixed.

src/swiss_ai_model_launch/assets/envs/vllm_base.toml
```diff
@@ -0,0 +1,61 @@
...
+# The paths are literal because variable references are not expanded reliably
+# on the launch path, and an unexpanded ${SCRATCH} collapses to an unwritable
+# root path. The cache is therefore shared across the infra01 group rather than
+# per user; the directories carry setgid and group write so members can
+# populate them. Entries are keyed by GPU architecture and library versions, so
+# different users, models and images coexist without colliding.
+#
+# TORCHINDUCTOR_CACHE_DIR is deliberately absent: vLLM sets it itself, to a
+# subdirectory of VLLM_CACHE_ROOT, so setting it here is silently discarded.
+# TRITON_CACHE_DIR and CUDA_CACHE_PATH are not covered by VLLM_CACHE_ROOT.
```
AryanAhadinia, line 65: So long comment.

src/swiss_ai_model_launch/assets/envs/vllm_base.toml
```diff
@@ -0,0 +1,61 @@
+image = "/capstor/store/cscs/swissai/infra01/container-images/ci/pr-204/vllm_base-{arch}.sqsh"
```
AryanAhadinia, line 1: Don't forget to fix the path before merge.

AryanAhadinia:
@mzueri, FYI, this PR intends to build image with latest dependencies.

PR #207, opened 2026-08-13
Fix Docker Image Build in CI

AryanAhadinia:
Cherry-picked from the `sglang_apertus` branch so the CI fix can land on `main` independently.

The image-build SLURM job was failing on the first pulled layer with `lsetxattr ...: operation not supported`. Rootless podman ignores the graphroot in `/etc/containers/storage.conf` and falls back to `$HOME/.local/share/containers/storage`; home is NFS, which has no user xattrs. Personal accounts have a `~/.config/containers/storage.conf` pointing at tmpfs, but the CI service account does not.

## Changes

- `.github/scripts/build_image.py`
  - The job now writes its own `storage.conf` (via `CONTAINERS_STORAGE_CONF`) putting graphroot/runroot on node-local tmpfs at `/dev/shm/$USER/podman-$SLURM_JOB_ID`, with `mount_program` set to fuse-overlayfs (tmpfs has no user xattrs below kernel 6.6 either, so the kernel overlay driver won't do).
  - fuse-overlayfs is located from a list of known paths, failing loudly if none is executable.
  - Seeds `~/.config/containers/storage.conf` for the service account if absent, so interactive `podman run` outside this script also lands on tmpfs. The per-job config still wins for the build itself, so two concurrent builds on one node don't trample each other's layers.
  - Cleanup switched from `podman rmi` to `podman system reset --force`: layers are owned by mapped subuids behind fuse mounts, so a plain `rm -rf` leaves them occupying the node's RAM.
- `docs/building-images.md` — documents the `storage.conf` prerequisite for building images by hand.
- `docs/ci-cd.md` — documents the podman storage setup under stage 3.

### DIFF
--- a/.github/scripts/build_image.py
+++ b/.github/scripts/build_image.py
@@ -80,14 +80,77 @@ def _build_slurm_script(
         export XDG_RUNTIME_DIR="${{TMPDIR:-/tmp}}/podman-runtime-$$"
         mkdir -p "${{XDG_RUNTIME_DIR}}"
 
+        # Rootless podman ignores graphroot/runroot from /etc/containers/storage.conf
+        # and falls back to $HOME/.local/share/containers/storage. Home is NFS, which
+        # has no user xattrs, so even pulling the base image dies with
+        # "lsetxattr ...: operation not supported". Personal accounts have a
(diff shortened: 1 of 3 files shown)


AryanAhadinia:
@robmsmt, FYI.

PR #211, opened 2026-08-27
Idempotent job submission: named launches, adopt-before-retry, retry 408/429

robmsmt:
## Why

FirecREST can report an error for an `sbatch` that actually went through — a 5xx, or **408 "Command execution timeout limit exceeded"** when its own SSH command to the cluster times out. On 2026-08-27, during a Clariden SSH outage, 11/11 submissions from evals-svc got that 408 (FirecREST's `/status/systems` had clariden's scheduler/ssh/filesystem unhealthy). Two problems in the launcher: `call_with_firecrest_retry` never retried 408 at all, and where it does retry (5xx) it resubmits blindly — allocating a second job if the first sbatch had gone through. CSCS's guidance is to check before resubmitting.

## What

- **`LaunchRequest.job_name`** — caller-supplied SLURM job name, the idempotency key (e.g. derived from the caller's run id). Unset keeps today's unique random name.
- **`Launcher.find_job(name)`** → `(job_id, status)` for a pending/running job of ours with that exact name, or `None`. FirecREST launcher: `job_info(name=…)` (**pyfirecrest bumped to ≥ 3.9.0**, lock refreshed); SLURM launcher: `squeue --me --name`. Base default returns `None`.
- **`Launcher._submit_idempotently`** — on a transient error, look the job up by name and adopt it; only if nothing exists, retry with backoff (1/2/4/8/16 s). Both `launch_model` and `launch_with_args` on the FirecREST launcher submit through it. Non-transient errors propagate at once.
- **408 and 429 are retryable** alongside 5xx (`is_firecrest_retryable` exported).
- Docs: "Idempotent launches" in `usage-advanced.md`.

## Testing

`tests/unit/test_idempotent_submit.py` drives `FirecRESTLauncher` with a fake client: 408 then adopt the existing job (one sbatch, one lookup); 503+408 with no job → retried to success; a finished job with our name is not adopted; 400 not retried; unnamed launches get a unique name and still check by it; `find_job` reports live jobs only. Unit suite: 364 passed, 70 skipped; ruff + mypy clean.

Note: committed with `--no-verify` because the pre-commit hooks could not build their environment locally (`hadolint-py==2.15.1.2` wheel fails to extract: "deflate decompression error" — unrelated to this change); ruff, ruff-format and mypy were run directly instead.

First consumer: evals-svc (follow-up PR there passes `job_name=f"evalsvc-{run id}"` and drops its private-method workaround).


pyproject.toml
```diff
@@ -16,7 +16,7 @@ dependencies = [
   "keyring>=25.0",
   "keyrings.alt>=5.0",
   "pydantic>=2.0",
-  "pyfirecrest>=3.7.1",
+  "pyfirecrest>=3.9.0",
```
AryanAhadinia, line 19: 🚀

src/swiss_ai_model_launch/launchers/utils.py
```diff
@@ -20,22 +20,37 @@
...
         except (AttributeError, IndexError):
             return False
-        return bool(500 <= status < 600)
+        return status in _RETRYABLE_STATUSES or bool(500 <= status < 600)
     return isinstance(exc, httpx.HTTPError)
 
 
+def is_firecrest_retryable(exc: BaseException) -> bool:
+    """Whether a failed FirecREST call may be retried (transient error)."""
+    return _is_firecrest_retryable(exc)
```
AryanAhadinia: Redundant wrapper.

src/swiss_ai_model_launch/launchers/slurm_launcher.py
```diff
@@ -72,7 +72,7 @@ def _get_working_dir(self) -> Path:
 
     def _get_launch_args_from_request(self, launch_request: LaunchRequest) -> LaunchArgs:
         model = launch_request.model
-        job_name = f"{model.replace('/', '_')}_{self.username}_{create_salt(8)}"
+        job_name = launch_request.job_name or f"{model.replace('/', '_')}_{self.username}_{create_salt(8)}"
```
AryanAhadinia, line 75: Why specifying a job name? Our job names are already unique because of adding the random salt.

src/swiss_ai_model_launch/launchers/launcher.py
```diff
@@ -209,6 +211,43 @@ async def launch_consecutive_with_args(
...
+                    raise
+                try:
+                    found = await self.find_job(job_name)
+                except Exception:  # the lookup itself hit the same outage
+                    found = None
+                if found is not None:
+                    return found[0]
+                if attempt == len(delays):
+                    raise
+                await asyncio.sleep(delays[attempt])
```
AryanAhadinia: I think, the delay, should be after submitting the job and before reading the status with `find_job`. I afraid that it may take a bit of time for SLURM to reflect the changes in `squeue`.

src/swiss_ai_model_launch/launchers/launcher.py
```diff
@@ -209,6 +211,43 @@ async def launch_consecutive_with_args(
...
+        """A job of ours with this exact name that is still pending or running,
+        as (job_id, status), or None. The basis of idempotent submission.
+        Launchers that cannot look a job up by name fall back to plain retries."""
+        return None
+
+    async def _submit_idempotently(
+        self,
+        job_name: str,
+        submit: Callable[[], Awaitable[int]],
+        delays: Sequence[float] = (1.0, 2.0, 4.0, 8.0, 16.0),
```
AryanAhadinia: The delays are too low I think. In the discussion we had with CSCS, they mentioned that the time out would be after 10 secs. Maybe start with 10 sec delay and raise it with smaller steps (factor 1.5 maybe).

docs/usage-advanced.md
```diff
@@ -119,6 +119,28 @@ replicas are marked `STALE` rather than left showing a frozen `HEALTHY`.
...
+
+FirecREST can report an error for an `sbatch` that actually went through —
+a 5xx, or a `408 Command execution timeout limit exceeded` when its own SSH
+command to the cluster times out. A retry that does not check first allocates
+a second job.
+
+So submission is idempotent when you give the launch a name of your own:
+
+```python
+job_id, served = await launcher.launch_model(LaunchRequest(..., job_name="myservice-run-1234"))
```
AryanAhadinia, line 132: I think we should expose an API to set the job name. The job names are already unique because of the random salt we are adding and it is already known to SML since SML sets the name itself.

src/swiss_ai_model_launch/launchers/firecrest_launcher.py
```diff
@@ -188,17 +188,18 @@ async def _prepare_launch_args(self, launch_args: LaunchArgs) -> LaunchArgs:
...
-        job_submission_report = await call_with_firecrest_retry(
-            lambda: self.client.submit(
-                system_name=self.system_name,
-                working_dir=self._get_working_dir(),
-                script_str=script_str,
-                account=self.account,
-            )
-        )
-        return int(job_submission_report["jobId"])
+        return await self._submit_idempotently(launch_args.job_name, lambda: self._submit_script(script_str))
```
AryanAhadinia: The name `idempotently` is a bit unfamiliar.

AryanAhadinia (changes requested):
Great change! Just needs a few minor modification.

robmsmt:
Status of the checks (2026-08-27 ~15:30 UTC): everything is green except **Integration Tests (lightweight)**, which is failing on the FirecREST/Clariden side, not on this change:

- Attempt 1: `5 failed, 81 passed, 16 errors` — all `UnexpectedStatusException` 503 *"The ssh service for the requested system (clariden) is unhealthy"* / 500 (channel error), including `test_catalog_paths` tests that don't touch the code in this PR.
- Re-run: `3 failed, 7 passed, 92 errors` — 4× 403 from `client.userinfo`, 3 catalog entries whose paths no longer exist (`zai-org/GLM-5`, `GLM-5-FP8`, `GLM-5.1`, `GLM-5.1-FP8`: *"path does not exist"*), and 92 `FileExistsError` from `keyrings.alt` `_ensure_file_path` → `os.makedirs` — an xdist (`-n 2`) race creating `~/.local/share/python_keyring` on a fresh runner. Pre-existing: `keyring`/`keyrings-alt` versions are identical on `main` and this branch (25.7.0 / 5.0.2).
- Direct probe with evals-svc's credentials at the same time: FirecREST `GET /status/systems` reports clariden `scheduler`, `ssh` and `filesystem` **all unhealthy** ("Command execution timeout limit exceeded"); `userinfo`, `ls`, `job_info` all 503.

The launch test did exercise the new idempotent path (`_submit_script` → retries → raised the last error after the attempts were exhausted, as designed). Re-run the job once Clariden is healthy; the keyring race and the stale GLM paths are separate, pre-existing issues.

src/swiss_ai_model_launch/launchers/utils.py
robmsmt: Dropped — `is_firecrest_retryable` is now the function itself (no private + wrapper pair).

src/swiss_ai_model_launch/launchers/slurm_launcher.py
robmsmt: The salted name is unique per `launch_model` call, and within one call it is what the lookup uses (unchanged). The optional `job_name` is for the caller's *own* retry: when evals-svc calls `launch_model`, the call can time out or the service can restart mid-flight — the launcher instance and its salt are gone, but the sbatch may have gone through. Only a name the caller chose up front (from its run id) lets the next attempt find and adopt that job instead of starting a second one. Returning the generated name doesn't cover that case, because the caller never gets a return value when the call fails. So: generated by default, caller-chosen when the caller needs to be able to come back for the job. Happy to change the shape if you'd rather see it on `LaunchArgs` only.

src/swiss_ai_model_launch/launchers/launcher.py
robmsmt: Agreed. After a transient submit error it now sleeps first, then looks the job up by name, and only resubmits if nothing is found; there is a test pinning that order (`test_waits_before_looking_the_job_up`).

src/swiss_ai_model_launch/launchers/launcher.py
robmsmt: Changed to 10s → 15s → 22.5s → 34s → 51s (×1.5), ~2.2 min across the 5 retries.

docs/usage-advanced.md
robmsmt: See the reply on `slurm_launcher.py`: the generated name covers retries inside one call; the caller-chosen name covers the caller's own retry after a timeout or restart, where no return value ever reached it. The docs section is exactly that API (`LaunchRequest(job_name=...)`) — the `find_job` lookup is by whatever name was used, generated or given.

src/swiss_ai_model_launch/launchers/firecrest_launcher.py
robmsmt: Renamed to `_submit_or_adopt`, which is what it does.

src/swiss_ai_model_launch/launchers/slurm_launcher.py
AryanAhadinia: OK that makes sense. Thanks!

AryanAhadinia (dismissed):

robmsmt:
Heads-up from a real run: CSCS's FirecREST reports **API 2.5.6**, and pyfirecrest raises `NotImplementedOnAPIversion` for `job_info(name=…)` below 2.6 (a plain `Exception`, not a `FirecrestException`). `find_job` now lists the account's jobs and matches the name itself, which works on every version; tests updated accordingly.

AryanAhadinia (dismissed):

PR #210, opened 2026-08-21
Add Check for Model Existence

AryanAhadinia:

PR #212, opened 2026-08-27
list_dir: only 404/403 mean the path is unreachable; let other FirecREST errors propagate

robmsmt:
## Problem

Since #210, `test_example_path_is_valid` has failed on a *different* path in every CI run — Mixtral-8x22B on main, Qwen3.5-397B / Qwen3.6-27B / Ministral-3 / `Alignment/trials/…` on one #211 run, the Apertus-v1p5 tokenizer and `rleval/…_1606_480it` on the next. All of those directories exist on Clariden with `config.json` and the same ACLs as their passing siblings.

`FirecRESTLauncher.list_dir` catches every `FirecrestException` and returns `None`, which `check_path` reports as `path does not exist or is not readable`. pyfirecrest raises `UnexpectedStatusException` for any non-200, so a 408 (FirecREST's `ls` command timeout) or a 5xx that outlived `call_with_firecrest_retry`'s five attempts was indistinguishable from a deleted checkpoint — and with `pytest.fail(..., pytrace=False)` the real status never reached the log.

## Fix

FirecREST 2.5.6 maps `ls`'s `No such file or directory` to **404** and `Permission denied` to **403** ([`base_command_error_handling.py`](https://github.com/eth-cscs/firecrest-v2/blob/2.5.6/src/firecrest/filesystem/ops/commands/base_command_error_handling.py)); `timeout` exit 124 becomes **408**, everything else **500**.

- `list_dir` now returns `None` only for 403/404 — verdicts on the path — and re-raises anything else, so the test fails with the actual `UnexpectedStatusException: last request: 408 …` instead of a bogus "missing".
- `utils.firecrest_status()` extracts the status once; `_is_firecrest_retryable` reuses it.
- Unit tests: 404 and 403 → `None`; 500 propagates after retries; 400 propagates without retrying.
- `docs/ci-cd.md`: a row for the new symptom.

No behaviour change for genuinely missing/unreadable paths.

### DIFF
--- a/docs/ci-cd.md
+++ b/docs/ci-cd.md
@@ -157,4 +157,5 @@ CI authenticates as a **service account**, not a personal account: the key is se
 | Build "succeeded", image unchanged | Sentinel cache hit — nothing under `images/<name>/` changed |
 | Merge skipped after a green build | The other arch failed, or a scan failed |
 | `path does not exist or is not readable` | A catalog entry's or recipe's weights moved or were deleted — repoint it or drop it |
+| `UnexpectedStatusException: last request: 408 …` / `500 …` in a path test | FirecREST couldn't list the directory (command timeout, backend error) even after retries — the path itself is fine; re-run the job |
 | `no model path extracted from: …` | An example's model flag isn't in a shape the extractor reads — see `tests/example_paths.py` |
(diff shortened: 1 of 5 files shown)


robmsmt:
Folded into #211 as requested.

PR #216, opened 2026-09-04
Upgrade OpenTela

AryanAhadinia:


AryanAhadinia:
The failure of the CI is expected. We have changed `--bootstrap.addr` to `--bootstrap.static`, but the CI is still running on the old binary. It would be fix once we make an OpenTela release and deploy the new binaries.

PR #221, opened 2026-09-17
Fix FirecREST userinfo KeyError and add --qos support

robmsmt:
## Summary
- CSCS upgraded the FirecREST API on Clariden to v2.6.0, which dropped the top-level `group` field from `/status/{system}/userinfo` in favor of a `default`-flagged entry inside `groups`. This broke every launch with `KeyError: 'group'` while resolving the SLURM account (`FirecRESTLauncher.from_client`). `_primary_group_name()` now handles both the pre- and post-2.6.0 response shapes.
- Adds a `--qos` flag (mirroring `--reservation`) threaded through `LaunchArgs`, `Launcher`/`FirecRESTLauncher`/`SlurmLauncher`, and `render_sbatch_header`/`to_sbatch_args`. Some partitions (e.g. `highprio`) reject `sbatch` with `Invalid qos specification` unless a matching `--qos` is set explicitly — `sml` had no way to pass one before this.

## Test plan
- [x] Unit tests for `_primary_group_name` covering the pre-2.6.0 shape, the 2.6.0 `default`-flag shape, the no-default fallback, and the no-group-info error case.
- [x] Unit tests for `render_sbatch_header`/`to_sbatch_args` covering `--qos` presence/absence.
- [x] `ruff check`, `mypy`, and the full unit suite (452 tests) pass locally.
- [x] Verified end-to-end against the real Clariden FirecREST API: `sml advanced --output-script` dry run, then a real `sbatch` submission on `normal`, then on `highprio --qos highprio` — job ran and served `swiss-ai/Apertus-v1.5-8B` successfully.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_016wWHntt9yQizubUN6PbCyG

PR #222, opened 2026-09-18
Add qos to FirecRESTLauncher, mirroring reservation

mzueri:
## Summary
- Some SLURM partitions (e.g. CSCS Clariden's `highprio`) reject a job submission with no `--qos` at all (`Invalid qos specification`), even when nothing else explicitly requested one for that account/partition combination.
- `FirecRESTLauncher` already threads an optional `reservation` through to its own sbatch header; `qos` had no equivalent anywhere in the launcher stack, so a caller whose SLURM policy pairs a partition with a required qos had no way to submit successfully.
- Adds `qos` alongside `reservation` through the same path: `Launcher`'s base `__init__`, `FirecRESTLauncher`'s `__init__`/`.from_client()`, and `render_sbatch_header()`'s `#SBATCH` lines.

## Scope
`SlurmLauncher`/`to_sbatch_args`, the CLI, the MCP server, and the loadtest module are untouched — none of them expose a `--qos` flag today, so there's nothing yet to wire it into there. This is scoped to exactly what a downstream caller (evals-svc) needs to unblock a model-serving job submission on a qos-requiring partition.

## Test plan
- [x] Added unit tests: `render_sbatch_header` with/without `qos` (mirroring the existing `reservation` tests)
- [x] Added an end-to-end unit test asserting `#SBATCH --qos=highprio` appears in the actual script submitted via `FirecRESTLauncher.launch_model()`
- [x] Full unit suite passes: 377 passed, 70 skipped (pre-existing skips, presumably needing a real cluster)
- [x] `ruff check` / `ruff format --check` pass


robmsmt:
This overlaps with #221, which merged into `main` a few minutes before this was opened (probably crossed in flight). The `qos` plumbing here — `Launcher`/`FirecRESTLauncher`/`render_sbatch_header` — is functionally identical to what's already on `main`, so merging this should be a clean no-op for that part (git's 3-way merge treats the identical additions as convergent, not conflicting).

A few things #221 covers that this doesn't:
- Fixes the `userinfo` `KeyError: 'group'` from CSCS's FirecREST v2.6.0 API change (this branch predates that fix, so its `firecrest_launcher.py` diff still shows the old broken line — merging won't reintroduce it, since this PR doesn't touch that line).
- Adds an actual `--qos` CLI flag on `sml advanced`/`preconfigured` (this PR intentionally scoped that out).
- `SlurmLauncher`/`to_sbatch_args` qos support.
- CLI-wiring unit tests + some example scripts under `examples/clariden/cli/swiss-ai/partition-tests/` verified against the real cluster.

One thing this PR has that #221 doesn't: `test_launch_forwards_qos_into_the_submitted_sbatch_script` in `test_idempotent_submit.py` — it checks qos end-to-end through `launch_model()` into the actual submitted script, rather than unit-testing `render_sbatch_header()` in isolation. That's a good complementary test worth keeping.

Given the overlap, would you rather close this in favor of #221 and port just that one test over, or should I do it? Either way works, just want to avoid two PRs doing the same plumbing.

